PR #51, opened 2026-04-15
Fix datetime TZ rendering and add Select option validation

derek-miller:
## Summary

Two fixes to entity behavior introduced in the current unreleased cycle, plus a host-mode ESPHome test fixture to exercise them.

- **Datetime render in project-local time.** The wire already carries a UTC epoch, but the Lua driver was reading with ``os.date("!*t", ...)`` (UTC) and writing with ``os.time{...}`` (local). The result was a TZ-shifted round-trip (saw +5h on CDT). Dropped the ``!`` on the read side so both directions use the project's local TZ. Matches ``C4:GetTimeZone()`` and how other C4 variables present datetime values.
- **Select option validation.** Writing an unknown option to a Select variable left the stale bogus value in C4 because the device silently ignored the command. The variable write handler now validates against the entity's ``options`` list; invalid values log an error and revert the variable. The ``Set Select`` programming command does the same check.
- **``test/dummy_inputs.yaml``.** Host-mode ESPHome config exposing Select, Event, Date, Time, and Datetime entities plus API services to change each from the device side, so both directions (C4→device and device→C4) can be driven from a test harness.

## Test plan

- [x] Select: C4→device via ``Set Select`` command, device→C4 via ``device_set_mode``, invalid option via variable write, invalid option via command
- [x] Event: fire each of press / double_press / long_press from the device and confirm C4 event fires + ``Test Button Last Event`` variable tracks the latest
- [x] Date: C4→device and device→C4 round-trip, invalid format logs parse error
- [x] Time: C4→device and device→C4 round-trip, invalid format logs parse error
- [x] Datetime: C4→device in local TZ → device stores matching UTC epoch, device→C4 UTC epoch → variable displays in local TZ, invalid format logs parse error

Bidirectional tests run against ``dummy_inputs.yaml`` on a host machine in ``America/Chicago`` and an ESPHome driver instance on the controller pointing at it.

### DIFF
--- a/src/esphome/entities/datetime.lua
+++ b/src/esphome/entities/datetime.lua
@@ -30,8 +30,10 @@ function DateTimeEntity:updated(entity, state)
     return
   end
 
+  -- The wire uses UTC epoch; format in the project's local time (os.date defaults to local)
+  -- so the variable round-trips with os.time on write.
   local epochSeconds = state.epoch_seconds or 0
-  local t = os.date("!*t", epochSeconds)
+  local t = os.date("*t", epochSeconds)
   local formatted = string.format("%04d-%02d-%02d %02d:%02d:%02d", t.year, t.month, t.day, t.hour, t.min, t.sec)
   values:update(entity.name, formatted, "STRING", function(newValue)
     local year, month, day, hour, minute, second = (newValue or ""):match(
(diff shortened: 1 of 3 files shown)


derek-miller:
@svc-finitelabs can you review this? Approval requested

derek-miller:
Good question. Checked the DriverWorks docs for ``OnVariableChanged``:

> OnVariableChanged is NOT called on a driver when it changes its own variable's value.

So the ``C4:SetVariable(entity.name, currentState)`` inside the OVC callback cannot re-enter OVC on that same driver - the runtime suppresses it at the source. That makes the revert safe without a guard or deferred timer.

Agreed the ``make init`` / ``package-lock.json`` churn from the rename-in-clone is local build noise, not something to chase in this PR.

PR #58, opened 2026-04-29
Add brightness, color, and scenes to ESPHome lights

derek-miller:
## Summary

ESPHome lights are no longer treated as on/off only. The driver now
detects each entity's capabilities at runtime and exposes the appropriate
controls in Composer and Navigator.

- **Brightness / dimming** with smooth ramps, preset and "Default On"
  mode handling, click and hold ramp rates, and hold-to-dim button
  control on virtual buttons and the On/Off/Toggle button-link bindings
- **Color and color-temperature** for every ESPHome color mode (white,
  color-temperature, cold/warm white, RGB, RGBW, RGB+white,
  RGB+color-temperature, RGB+cold/warm white)
- **Advanced Lighting Scenes**: lights can be scene members with
  brightness and color steps, ramp/stop, and flash mode

The parent ESPHome driver also notifies entity sub-drivers when the
device disconnects so they reset state correctly on reconnect.

## Test plan

- [x] Eight dummy ESPHome lights covering every color mode (`test/dummy_lights.yaml`)
- [x] C4 → ESPHome: on/off, set level, brightness target with rate, color target (XY and CCT/Kelvin), button events, button-link DO_PUSH/DO_RELEASE/DO_CLICK
- [x] ESPHome → C4: brightness and color state propagation via aioesphomeapi
- [x] Disconnect / reconnect cycle (entity sub-drivers reset and re-discover)
- [x] Real Advanced Lighting Scenes agent → driver chain: scene created with member and element through the agent, activated, ESPHome state matches the configured brightness + color
- [x] Verified against a real ESPHome bulb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,18 @@
 [//]: # "- Removed"
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Added
+
+- Added brightness and dimming support to ESPHome lights with smooth ramping,
+  preset management, and hold-to-dim button control
+- Added color and color-temperature support to ESPHome lights for every ESPHome
+  color mode (white, color-temperature, cold/warm white, RGB, RGBW, and combined
+  RGB + white modes)
+- Added Advanced Lighting Scenes support to ESPHome lights so they can
+  participate in lighting scenes alongside other Control4 dimmers
+
 ## v20260418 - 2026-04-18
 
 ### Added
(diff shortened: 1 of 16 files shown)


derek-miller:
@svc-finitelabs please review

derek-miller:
Thanks for the catch — fixed in dfe5812. SET_COLOR_TARGET now falls back to the persisted default color rate (set via UPDATE_COLOR_RATE_DEFAULT) when no per-command rate is provided, matching the brightness path.

derek-miller:
@svc-finitelabs your approval is required

PR #62, opened 2026-05-24
DRV-43: replace BT proxy hardware rec with affiliate-linked Seeed W5500

derek-miller:
## Summary

Replaces the standalone **XIAO ESP32C6** recommendation in the Bluetooth Proxy Configuration Guide with the **XIAO W5500 Ethernet Adapter** (paired with a XIAO ESP32S3 Plus), using Seeed's affiliate-tracked URL so purchases through the docs support driver development.

- Swap C6 → W5500, using the dashboard-generated tracking format `?sensecap_affiliate=sikqXSu&referring_service=link`
- Keep the Olimex ESP32-POE / ESP32-POE-ISO boards as a plain (non-affiliate) alternative
- Add an FTC-style affiliate disclosure under the Hardware list
- Regenerate `README.md` from the doc source to match

Only `index.md` (the source) and the generated `README.md` change. The W5500 affiliate link is the exact URL generated from the Seeed affiliate dashboard; the Olimex links stay plain since the Seeed code doesn't apply to a different vendor.

Resolves [DRV-43](https://youtrack.dmiller.me/issue/DRV-43).

## Notes

- This reaches controllers only on the next `make build` + release (docs are bundled into `esphome.c4z`); no changelog entry added since this is a docs/hardware-rec change, not a driver behavior change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,11 +664,13 @@ BLE devices into Control4.
 The following POE-powered Bluetooth proxies are excellent choices with 4 active
 connection slots:
 
-- [Seeed Studio XIAO ESP32C6](https://www.seeedstudio.com/Seeed-Studio-XIAO-ESP32C6-p-5884.html)
-  with POE expansion board
+- [Seeed Studio XIAO W5500 Ethernet Adapter](https://www.seeedstudio.com/XIAO-W5500-Ethernet-Adapter-p-6472.html?sensecap_affiliate=sikqXSu&referring_service=link)
 - [Olimex ESP32-POE](https://www.olimex.com/Products/IoT/ESP32/ESP32-POE/open-source-hardware)
   or
   [ESP32-POE-ISO](https://www.olimex.com/Products/IoT/ESP32/ESP32-POE-ISO/open-source-hardware)
+
+_Some hardware links above are affiliate links. Purchases through them help
+support driver development at no extra cost to you._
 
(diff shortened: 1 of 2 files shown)


derek-miller:
@svc-finitelabs can you review? Approval requested

derek-miller:
Thanks for the careful review. Both points trace back to the product listing being misread (by me as well, originally). The linked SKU is the **XIAO W5500 Ethernet Adapter V1.2 (ESPHome Ready-Made Project)**, which is a complete, assembled unit rather than a bare carrier board:

**1. "Needs a XIAO ESP32S3 Plus":** The board *includes* it. Seeed describes it as "a compact PoE development board featuring the XIAO ESP32S3 Plus," shipping ESPHome-ready with enclosure and antenna. It's a working BLE proxy out of the box, so the bare link is correct, and a "paired with..." note would actually be inaccurate.

**2. PoE:** It has an integrated, isolated PoE module (TPS563201-based 5V conversion) and is powered directly from a PoE switch, no external injector. So the "POE-powered" header holds, same as the Olimex boards.

No doc changes needed on those two, but appreciate the catch. It's exactly the kind of thing that matters on buyer-facing docs.

derek-miller:
@svc-finitelabs see previous comment

derek-miller:
@svc-finitelabs you need to approve

PR #65, opened 2026-07-11
Restore dynamic bindings in OnDriverInit so connections survive Director restarts

derek-miller:
## Problem

Connections to the Bluetooth Coordinator input on ESPHome proxies disappeared after controller reboots / Director restarts (observed repeatedly on a production system; its project DB contained six stale generations of the lost connection rows, one per past rebind/loss cycle).

## Root cause

Director stores a connection as a child of the **provider's** binding row (`bound_consumers.parent_id -> bindings.id` in `project.db`) and never persists consumer-side dynamic bindings. At startup, Director resolves stored connections **before `OnDriverLateInit`**:

- Provider-side dynamic bindings self-heal: `AddDynamicBinding` on a provider re-scans and re-attaches its stored consumers at any time.
- Consumer-side dynamic bindings have no re-attach hook. If the consumer binding does not exist when the resolution pass runs, the connection is silently and permanently dropped.

Our drivers restored dynamic bindings in `OnDriverLateInit`, which is after resolution, so the coordinator connection (the only consumer-side dynamic binding in the ecosystem) was lost on **every** Director restart. Reproduced 1-for-1 on a test controller.

This matches the DriverWorks docs: the `AddDynamicBinding` restore example runs at script load, `PersistGetValue` is documented usable before LateInit, and the safe-usage table notes variables must be added in `OnDriverInit` while `AddEvent` must wait for `OnDriverLateInit`.

## Changes

- Restore dynamic bindings and variables in `OnDriverInit` in all drivers (esphome, coordinator, climate, yale; govee/bthome/switchbot already did this).
- Move event restore from `OnDriverInit` to `OnDriverLateInit` in govee/bthome/switchbot (`C4:AddEvent` is unavailable earlier; opposite violation of the same lifecycle table).
- Document the lifecycle constraints on `restoreBindings()`, `restoreValues()`, and `restoreEvents()` in the lib so callsites can stay one-line.

## Validation (test controller)

- Baseline (previous code): coordinator connections to 2 proxies dropped on the first Director restart.
- With the fix: connections to 3 proxies, including one with a device ID **above** the coordinator's (worst-case load order), survived 5 Director restarts (3 with an instrumented build, 2 with this exact build).
- Provider-side connections (lights/fans/relays/BT child devices) unaffected throughout.

## Follow-ups (separate PRs)

- Guard the `restoreBindings()` prune when persist reads empty.
- Register coordinator-binding RFP/OBC handlers at init instead of on capability discovery.
- Coordinator: accept `PROXY_CONNECTED` from unknown-but-bound proxies.

After updating, any previously lost connections need one manual rebind; they will persist from then on. The same latent bug exists in control4-tplink and control4-home-connect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
 
 ### Fixed
 
+- Fixed Bluetooth Coordinator connections to ESPHome proxies (and other
+  dynamically created connections) disappearing after a controller reboot or
+  Director restart. Dynamic bindings are now restored early enough in driver
+  startup for Director to reconnect them.
 - Declare ESPHome light hardware capabilities (dimming, color, color
   temperature) conservatively in the static baseline and enable them at runtime
   from the entity's discovered color modes. A full static baseline advertised
--- a/README.md
+++ b/README.md
@@ -880,6 +880,10 @@ can file an issue on GitHub:
 
 ### Fixed
 
+- Fixed Bluetooth Coordinator connections to ESPHome proxies (and other
+  dynamically created connections) disappearing after a controller reboot or
+  Director restart. Dynamic bindings are now restored early enough in driver
+  startup for Director to reconnect them.
 - Declare ESPHome light hardware capabilities (dimming, color, color
   temperature) conservatively in the static baseline and enable them at runtime
   from the entity's discovered color modes. A full static baseline advertised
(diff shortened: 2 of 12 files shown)


derek-miller:
@svc-finitelabs can you review and approve if it looks good?

PR #70, opened 2026-07-28
Expose temperature and humidity sensors as Control4 value connections

derek-miller:
ESPHome `sensor` entities were published only as name-keyed **NUMBER variables** — `device_class` was discarded, so a DHT22's Temperature and Humidity couldn't be bound to a thermostat or anything else expecting a value class. `sensor` was the only entity type that created no dynamic binding.

## Change

Sensors whose `device_class` maps to a C4 value class also get a `CONTROL` provider binding:

```lua
local SENSOR_BINDINGS = {
  temperature = { bindingClass = "TEMPERATURE_VALUE", scale = "CELSIUS" },
  humidity    = { bindingClass = "HUMIDITY_VALUE",    scale = "PERCENT" },
}
```

- `VALUE_CHANGED` is pushed only when the value actually changes; consumers are seeded on bind (`OBC`) and `GET_VALUE` is answered from cache.
- The reported `SCALE` honors the sensor's declared `unit_of_measurement` — the native API's only unit signal for sensor entities — matched case-insensitively with whitespace/degree signs stripped, mapping C/F/K to `CELSIUS`/`FAHRENHEIT`/`KELVIN` (Celsius default). Consumers like `esphome_climate` convert based on the scale.
- Mirrors the provider pattern in `esphome_bthome`; more mappings are one table row each.

Unmapped sensors are unchanged and stay variable-only.

## Verification

- **Live on production:** `TEMPERATURE_VALUE` (id=14) / `HUMIDITY_VALUE` (id=13) bindings created and visible on a ratgdo32; sensor values flowing through the same path.
- Unit-normalization matching covered by a small test harness (°F, F, " °f ", °C, K, °K, %, empty, nil).
- Not yet exercised live: a Composer connection to a consumer, and the Fahrenheit/Kelvin branches (test devices report °C) — statically verified only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ## Unreleased
 
+### Added
+
+- Added `TEMPERATURE_VALUE` and `HUMIDITY_VALUE` connections for sensors with a
+  `temperature` or `humidity` device class, so they can be bound to thermostats
+  and other value consumers; the reported scale honors the sensor's declared
+  unit of measurement
+
 ### Fixed
 
 - Fixed all entity states appearing frozen in Control4 (covers stuck on
@@ -28,7 +35,7 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
   `object_id`, and log messages now identify entities as `type 'Name' (key=N)`.
 - Bumped the ESPHome native API version advertised in `HelloRequest` from 1.0 to
   1.14 so devices no longer log `using outdated API 1.0, update to 1.14+` on
-  every connection.
+  every connection
 
 ## v20260711 - 2026-07-11
 
--- a/README.md
+++ b/README.md
@@ -908,6 +908,13 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ## Unreleased
 
+### Added
+
+- Added `TEMPERATURE_VALUE` and `HUMIDITY_VALUE` connections for sensors with a
+  `temperature` or `humidity` device class, so they can be bound to thermostats
+  and other value consumers; the reported scale honors the sensor's declared
+  unit of measurement
+
 ### Fixed
 
 - Fixed all entity states appearing frozen in Control4 (covers stuck on
@@ -916,7 +923,7 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
   `object_id`, and log messages now identify entities as `type 'Name' (key=N)`.
 - Bumped the ESPHome native API version advertised in `HelloRequest` from 1.0 to
   1.14 so devices no longer log `using outdated API 1.0, update to 1.14+` on
-  every connection.
+  every connection
 
 ## v20260711 - 2026-07-11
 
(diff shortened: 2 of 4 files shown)


derek-miller:
@svc-finitelabs please review

derek-miller:
Both addressed in the updated commit:

1. **Per-entity scale** — implemented essentially as suggested: `getScale(entity, config)` maps `unit_of_measurement` of `°F`/`F` to `FAHRENHEIT` for `TEMPERATURE_VALUE` bindings, default otherwise, used at both `SendToProxy` sites. On spelling: went with `FAHRENHEIT` to pair with the existing `CELSIUS`/`PERCENT` convention (`esphome_bthome` precedent), and the in-repo consumer treats any non-`C` spelling as Fahrenheit. The stock-C4-consumer question stands unverified — if a stock consumer turns out to want `F`, it's a one-line table change.

2. **Change-gated notify** — `VALUE_CHANGED` now fires only when `values:update` reports an actual change. Consumers still get seeded on bind and can `GET_VALUE` on demand, which covers resync without the periodic-refresh chatter.

derek-miller:
Follow-up on the scale handling after checking the #72 schema: the new `TemperatureUnit` enum landed only on the **climate** and **water-heater** list responses — `ListEntitiesSensorResponse` is unchanged, so `unit_of_measurement` remains the only unit signal for sensor entities on every firmware version. Matching is now normalized (case-insensitive, whitespace and degree signs stripped) and maps `C`/`F`/`K` to `CELSIUS`/`FAHRENHEIT`/`KELVIN`, mirroring the enum's values. Honoring `temperature_unit` on the climate/water-heater path (where the wire is no longer guaranteed Celsius on 2026.7+) is a separate change gated on #72 merging.

derek-miller:
Latest review items addressed:

- **Restart reseed**: the push gate is now an in-memory last-pushed value per binding instead of the persisted `changed` return — identical repeats within a session still drop, and the first reading after a driver restart always goes out, closing the stale-consumer gap you flagged.
- **Kelvin consumer**: added the `K`/`KELVIN` branch to `getCelsiusFromParams` in `esphome_climate`, so the scale this PR makes reachable is also consumable (295 K → 21.9 °C instead of f2c's 146).
- **Locale nit**: degree/whitespace stripping now runs before `:lower()`.

The same persisted-gate restart flaw exists on main in cover contacts, Yale DoorSense, BTHome bindings, and the SwitchBot Bot relay — separate PR incoming for those.

derek-miller:
@svc-finitelabs please review

derek-miller:
Correction for the record: I wrote above that "the wire is no longer guaranteed Celsius on 2026.7+" for climate/water-heater. Checked the 2026.7.2 tag's C++ — `temperature_unit` exists in api.proto but nothing populates it at that version (zero references in api_connection.cpp / climate_traits / water_heater; the wiring is only on dev). So the wire remains Celsius on all released firmware, the field decodes as the proto3 default `CELSIUS`, and the climate-side follow-up is future-proofing for an unreleased ESPHome, not a current gap.

PR #71, opened 2026-07-28
Fix state updates dropped for ESPHome firmware that omits object_id

derek-miller:
Newer ESPHome firmware (observed with **2026.7.2**) no longer sends the deprecated `object_id` field in list entities responses. The state dispatch guard required it, so every state update was dropped as `Received state update for unknown entity` while the connection looked healthy — covers stuck on **"Unknown"**, sensor variables never created. Devices on older firmware still send the field and were unaffected.

## Fix — remove the `object_id` dependency entirely

- State dispatch requires only `entity_type`; entity identity is the numeric key, as it always was.
- Restart-button detection matches the entity **name** — same selectivity, since `object_id` was derived from the name (`device_class` can't be used: both *Restart* and *Safe mode boot* carry `"restart"`).
- All entity log sites use a shared `ESPHomeClient.describeEntity` helper producing `type 'Name' (key=N)` — display names read cleanly and duplicate names stay distinguishable by key.
- Empty-named climate/water-heater entities take the **device name** for their sub-driver display name (Home Assistant semantics for a device's main entity) instead of reconstructing it from `object_id`; the client now caches device info and exposes `getDeviceName()`, which `RefreshStatus` also uses for the `Name` value. Fallbacks use `IsEmpty` rather than `or`, since unset proto string fields decode as `""` (truthy in Lua).
- Fixes a pre-existing copy-paste typo in `text.lua`'s failure log (`number.%s`).

After this, no code path reads `object_id` — it survives only as wire-schema passthrough and a name-first fallback in one test utility. Behavior is identical on firmware that still sends the field.

## Verification

- **Live on production:** both affected ratgdo32 devices (2026.7.2) recovered the full entity set on the first refresh after deploy; the bound garage-door proxy went from `STATE = "Unknown"` to the correct state with no door movement; states render as `State update for sensor 'Temperature' (key=899752953) -> {"state":32.5}`; name-based restart detection picks the correct button; an unaffected 2026.3.1 device verified for regression.
- Every `log:*` call's format-specifier/argument arity audited programmatically — all match.
- `luajit` syntax checks pass; stylua/mdformat idempotent; `make preprocess gen-squishy update-xml package` exits 0; generated CHANGELOG/README formatting committed.
- The empty-name display path is not exercisable on the test devices (all entities named) — verified statically only.

## Notes

- Trivial CHANGELOG conflict with #70 (both touch `Unreleased`).
- The latent bug found during this investigation — `disconnect()` clears the callback registry without rejecting in-flight deferreds, silently stalling any awaiting chain — is tracked as **DRV-55**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed all entity states appearing frozen in Control4 (covers stuck on
+  "Unknown", sensor variables never updating) on ESPHome 2026.7+ firmware, which
+  no longer sends the deprecated `object_id` field. The driver no longer reads
+  `object_id`, and log messages now identify entities as `type 'Name' (key=N)`.
 - Bumped the ESPHome native API version advertised in `HelloRequest` from 1.0 to
   1.14 so devices no longer log `using outdated API 1.0, update to 1.14+` on
   every connection.
--- a/README.md
+++ b/README.md
@@ -910,6 +910,10 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed all entity states appearing frozen in Control4 (covers stuck on
+  "Unknown", sensor variables never updating) on ESPHome 2026.7+ firmware, which
+  no longer sends the deprecated `object_id` field. The driver no longer reads
+  `object_id`, and log messages now identify entities as `type 'Name' (key=N)`.
 - Bumped the ESPHome native API version advertised in `HelloRequest` from 1.0 to
   1.14 so devices no longer log `using outdated API 1.0, update to 1.14+` on
   every connection.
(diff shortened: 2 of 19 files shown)


derek-miller:
Review findings addressed in `1e7f2cb`: restored the `service.method` specifiers in all ten command-handler log sites per the inline suggestions (the conversion had collapsed both `%s.%s` pairs; my arity check counted lines, not occurrences, so double-specifier strings slipped through). Re-audited every log call in the driver programmatically afterward — all arities match. Generated CHANGELOG/README formatting is committed and `fmt-md`/`docs-readme` are idempotent locally against CI's output, and the changelog entry is tightened per review. PR description updated to match the final no-`object_id` shape, including a correction: the `return reject(err)` follow-up I'd listed was wrong — `reject` is the defined helper in `lib/utils.lua`.

derek-miller:
@svc-finitelabs I need your approval of everything looks good

PR #73, opened 2026-07-29
Re-notify bound consumers after a driver restart

derek-miller:
Four drivers deduplicate proxy notifications by comparing against the **persisted** variable value (`if values:update(...)`/`if changed`). Persistence survives driver restarts, so the first reading of a new session is suppressed whenever it matches the stored value — and the bound consumer keeps whatever state it had until the device actually changes state:

- `cover.lua` contact sensors — a garage-door proxy sits on **Unknown** after a driver update until the door physically moves (observed on production during today's debugging: the one reload that healed it did so only because the persisted values happened to restore as nil)
- `esphome_yale` DoorSense contact (the docstring even said "Persists state via the values lib so it survives reboots" — that persistence is precisely the bug)
- `esphome_bthome` sensor and contact bindings
- `esphome_switchbot` Bot relay

## Fix

Deduplicate against an **in-memory** last-notified value per binding instead. Identical repeats within a session still drop (preserving the v20260325/v20260318-era duplicate-notification fixes), and a driver restart clears the memory so the first reading always goes out. The persisted variables are untouched — `values:update` still runs unconditionally.

`binary_sensor` and `esphome_govee` push unconditionally and never had the flaw. The same fix for the new sensor value bindings is in #70.

## Verification

- `luajit` syntax checks pass; stylua/mdformat idempotent; `make preprocess gen-squishy update-xml package` exits 0.
- Statically verified only — exercising it requires a driver reload against a device whose state matches persistence, which is exactly the hard-to-reproduce condition; the cover case was observed in the wild today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed cover contacts, Yale DoorSense, BTHome bindings, and the SwitchBot Bot
+  relay leaving bound consumers stale after a driver restart or update until the
+  next state change
 - Fixed all entity states appearing frozen in Control4 (covers stuck on
   "Unknown", sensor variables never updating) on ESPHome 2026.7+ firmware, which
   no longer sends the deprecated `object_id` field. The driver no longer reads
--- a/README.md
+++ b/README.md
@@ -917,6 +917,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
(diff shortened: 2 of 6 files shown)


derek-miller:
@svc-finitelabs please review

derek-miller:
@svc-finitelabs please review (head is `f2d9142` — doc-comment placement fixed since the first request)

derek-miller:
@svc-finitelabs please review

derek-miller:
@svc-finitelabs the blank line at `esphome_bthome/driver.lua:632` is deleted — thanks for going to the LuaLS parser source to settle it. Please re-review, and if it's clean, submit a formal approving review so branch protection is satisfied.

PR #74, opened 2026-07-29
Re-notify SwitchBot consumers after a driver restart

derek-miller:
Follow-up to #73 for the remaining `esphome_switchbot` sites tracked in [DRV-56](https://youtrack.dmiller.me/issue/DRV-56):

- `setChannel1State` / `setChannel2State` gated `SendToProxy` on `values:update`'s persisted return
- `sendContactState` gated on persisted `previousState.contact_*`, so motion/contact/leak/tamper CONTACT_SENSOR consumers could sit on the wrong state indefinitely after a driver update (the persisted keys are only cleared on ESPHome rebind or driver reset, never on a plain restart)

All three now dedupe against an in-memory last-notified memo, recorded only after a confirmed send (the `cover.lua` pattern from #73). The Bot relay memo from #73 is aligned to the same record-after-confirmed-send ordering, closing the early-poison window noted in that review. The ESPHome rebind (`OBC[ESPHOME_BINDING]`) and `EC.Reset_Driver` paths call `clearNotifiedState()` alongside their persisted-state clears, so a forced resync still re-notifies (review finding).

Deliberately unchanged:

- `fireStateChangeEvent` (and the govee equivalent) still gates C4 programming events on persisted state — duplicate `leak_detected`/`motion_detected` events must not re-fire across a driver update, and the nil guard already prevents initial-load firing
- `contactButtonCount` press detection stays persisted for the same reason

Also fixes a stray `%s` in `sendContactState`'s trace format string, which raised a format error whenever Log Mode = Trace.

## Verification

`luajit` syntax check passes, stylua/mdformat idempotent, full package build exits 0. Statically verified — exercising it requires a driver restart against a device whose state matches persistence, the same constraint as #73.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Removed
 -->
 
+## Unreleased
+
+### Fixed
+
+- Fixed SwitchBot channel relays and contact sensors (motion, contact, leak,
+  tamper) leaving bound consumers stale after a driver restart until the next
+  state change
+
 ## v20260728 - 2026-07-28
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -906,6 +906,14 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Removed
 -->
 
+## Unreleased
+
+### Fixed
+
+- Fixed SwitchBot channel relays and contact sensors (motion, contact, leak,
+  tamper) leaving bound consumers stale after a driver restart until the next
+  state change
+
 ## v20260728 - 2026-07-28
 
 ### Added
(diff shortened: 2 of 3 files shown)


derek-miller:
@svc-finitelabs please review. Extends the #73 in-memory re-notify fix to the remaining switchbot sites per DRV-56 — pay particular attention to the record-after-confirmed-send ordering and the deliberately-unchanged persisted event gating in fireStateChangeEvent/contactButtonCount.

derek-miller:
@svc-finitelabs both your reviews' blocking finding is addressed at `e8659a3`: `clearNotifiedState()` (declared below all three memos, so the upvalue assignments bind correctly) now runs in `OBC[ESPHOME_BINDING]` and `EC.Reset_Driver` alongside their `persist:set("previousState", {})` clears, restoring the rebind/reset resync behavior. The same gap in the #73 drivers (bthome's `bindings:reset()`, yale, cover.lua) is being filed as its own ticket rather than expanding this PR. Please re-review, and approve if it now passes.

derek-miller:
@svc-finitelabs your re-review at `e8659a3` found nothing else to raise — please submit a formal GitHub approval review on this PR so the review requirement is satisfied.

PR #75, opened 2026-07-29
Reject in-flight requests on disconnect

derek-miller:
Fixes [DRV-55](https://youtrack.dmiller.me/issue/DRV-55), split out of #71.

`ESPHomeClient:disconnect()` cancelled every pending callback's timeout timer — the only mechanism that would have rejected a waiting request — and then discarded the registry. Any promise chain awaiting a response at disconnect time stalled silently forever: `RefreshStatus()` chains hung mid-refresh instead of reporting the failure, and BLE GATT operations never ran their cleanup handlers. Reachable in normal operation, since `disconnect()` runs on reconnect, refresh failure, and driver teardown.

## Fix

- `_registerCallback` stores the request's abort handler (`onTimeout`) on the registry entry.
- `disconnect()` swaps the registry out up front (so abort handlers can re-enter `disconnect()` or register new callbacks), then invokes each stored handler with a `"Disconnected"` reason under `pcall` as its final step — after the Bluetooth/user-services resets and socket close — so a rejection handler can safely start a reconnect (review finding).
- The 11 abort closures accept the optional reason (`err or "<original timeout message>"`), so timeouts keep their existing messages and disconnects report `Disconnected`.
- The handshake closures only run `checkHandshakeState`/`_hsState = ERROR` on a real timeout — a disconnect abort neither logs a bogus "Expected Noise state HANDSHAKE, actual nil" error nor resurrects the already-cleared state (review findings).
- Persistent subscription callbacks (ping/time handlers, state and BLE advertisement subscriptions — registered without a timeout) are dropped without invocation, exactly as before.
- `_btProxyInitDeferred` settles transitively: its chain rides on registry callbacks that now reject.

Noted, not changed: `subscribeBluetoothConnections` uses a raw `C4:SetTimer` (not the registry) for its initial-response wait; on disconnect it still settles, just up to 10s late. The `_registerCallback` key-collision case (overwrite strands the previous waiter) is being filed as its own ticket per review.

## Verification

`luajit` syntax check passes, stylua/mdformat idempotent, full package build exits 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed pending requests (refresh, Bluetooth GATT operations) hanging forever
+  when the connection dropped mid-request; they now fail immediately with a
+  "Disconnected" error
 - Fixed SwitchBot channel relays and contact sensors (motion, contact, leak,
   tamper) leaving bound consumers stale after a driver restart until the next
   state change
--- a/README.md
+++ b/README.md
@@ -910,6 +910,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed pending requests (refresh, Bluetooth GATT operations) hanging forever
+  when the connection dropped mid-request; they now fail immediately with a
+  "Disconnected" error
 - Fixed SwitchBot channel relays and contact sensors (motion, contact, leak,
   tamper) leaving bound consumers stale after a driver restart until the next
   state change
--- a/src/esphome/client.lua
+++ b/src/esphome/client.lua
@@ -42,6 +42,7 @@ local Indicator = {
 --- @class CallbackEntry
 --- @field callback CallbackFunction The callback function to invoke
 --- @field timer C4LuaTimer|nil Optional timeout timer for auto-unregistration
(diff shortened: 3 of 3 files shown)


derek-miller:
@svc-finitelabs please review. Fixes DRV-55 (your ticket): disconnect() now rejects in-flight request deferreds via stored abort handlers. Pay particular attention to re-entrancy (rejection handlers calling disconnect()/reconnect), the persistent-vs-request callback split, and whether any abort closure misbehaves when invoked mid-teardown.

derek-miller:
@svc-finitelabs all three findings addressed at `b287b1d`: (1+2) the handshake closures gate `checkHandshakeState` and the `_hsState = ERROR` write on `not err`, so a disconnect abort neither logs the bogus mismatch nor resurrects the cleared state; (3) the abort loop now runs as the final step of `disconnect()`, after the Bluetooth/user-services resets and `client:Close()`, with the registry still swapped up front — making the reconnect-from-handler claim actually true. The key-collision follow-up you flagged is being filed as its own ticket. Please re-review, and approve if it now passes.

derek-miller:
@svc-finitelabs your re-review at `b287b1d` confirmed all three findings addressed — please submit a formal GitHub approval review on this PR so the review requirement is satisfied.

derek-miller:
@svc-finitelabs the rebase onto merged #74 dismissed your approval. Head is now `cf5e08d` — identical code to the `b287b1d` you approved; the only delta is the changelog/README conflict resolution stacking this entry above #74's. Please re-review and submit a formal approval if it passes.

PR #78, opened 2026-07-29
Clear notify memos on driver reset

derek-miller:
Fixes [DRV-60](https://youtrack.dmiller.me/issue/DRV-60) — the reset-path gap flagged in the #74 review, plus one site the ticket underestimated.

The in-memory notify memos clear on a driver restart but survive `EC.Reset_Driver`, which recreates dynamic bindings and wipes stored values expecting the next reading to resync consumers. The memo suppressed that first notify. One correction to the review's scope: the main esphome driver **does** have a `Reset_Driver` (`drivers/esphome/driver.lua:644`) with `bindings:reset()` + `values:reset()`, so cover.lua's memo — and sensor.lua's `lastPushed` from #70, which the ticket missed entirely — were both affected too.

## Changes

- `cover.lua` / `sensor.lua`: export `clearNotifiedState()` (static, reassigns the file-local memo upvalue); the main driver's `Reset_Driver` calls both after `values:reset()`
- `esphome_bthome`: `lastNotified = {}` in its `Reset_Driver`'s local-state block
- `esphome_yale`: `lastNotifiedDoorStatus = nil` in its `Reset_Driver`

switchbot needs nothing here: #74 (now in this branch's base) already clears all three of its memos via `clearNotifiedState()` on both its reset and rebind paths. With that, every notify memo in the repo clears on reset: bthome, yale, switchbot ×3, and the two entity modules.

Rebind (`OBC[ESPHOME_BINDING]`) paths deliberately unchanged: bthome only resets `knownObjects` there (its per-binding OBC handlers re-push on bind, as does sensor.lua's `sendCachedValue`), and yale's only calls `resetConnectionState`. Clearing memos there would invent new rebind semantics rather than restore old ones. The reviewer-confirmed gap that cover/yale contacts have no bind-time seeder at all is pre-existing #73 behavior, tracked separately.

On the shared-helper suggestion: skipped for now — each memo is file-local next to its gate, and a shared lib would still need per-driver clear wiring. Can revisit if the pattern spreads further.

## Verification

`luajit` syntax on all changed files, stylua/mdformat idempotent, full package build exits 0. Pure additions, 25 insertions.

PR #80, opened 2026-07-29
Seed contact bindings with the last known state on bind

derek-miller:
Fixes [DRV-62](https://youtrack.dmiller.me/issue/DRV-62), the gap confirmed in the #78 review: cover's `cover_open_*`/`cover_closed_*` contacts and yale's `door` contact had no bind-time seeder, so a newly bound consumer sat stale until the physical state changed — with the notify memo suppressing the resync. Sensor values (`sendCachedValue` OBC), BTHome (per-binding OBC re-push), and SwitchBot (memo clear in its OBC) all already handle bind; these two were the odd ones out.

## Changes

- `cover.lua`: `registerContactSeeder(bindingId, valueName)` registers an OBC per contact binding that sends the stored `<name> Open`/`<name> Closed` value (`CLOSED`/`OPENED`) on `bIsBound`. `discovered()` now captures the contact binding ids and registers both seeders — `discovered()` runs on every connect, so the handlers exist every session.
- `esphome_yale`: `registerDoorContactSeeder()` sends `lastNotifiedDoorStatus` (this-session truth) falling back to the persisted `Door Status` value. Registered from **both** the `setDoorSenseConfigured(true)` create path and the `OnDriverLateInit` restore path — the restore sets `doorSenseConfigured` directly, which would otherwise early-return the create path and leave the session without a handler. The memo declaration moved above the seeder for upvalue capture (declaration order, per the #74 lesson).

Seeds bypass the memo and don't record it, matching bthome/sensor: worst case is one redundant idempotent notify on the next state dump.

Known tradeoff (same as the sensor seed, noted in the #73/#74 reviews): the persisted fallback can be stale if the device changed while the driver was down; the next real state dump corrects it because the memo comparison still sees the change.

## Verification

`luajit` syntax on both files, stylua/mdformat idempotent, full package build exits 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed cover contacts and Yale DoorSense sending nothing to a newly bound
+  consumer (or after a DoorSense drop and re-detect) until the state changed;
+  consumers are now seeded with the last known state on bind
 - Fixed the Reset Driver action leaving bound consumers stale until the next
   state change (sensor values, cover contacts, BTHome bindings, Yale DoorSense)
 - Fixed pending requests (refresh, Bluetooth GATT operations) hanging forever
--- a/README.md
+++ b/README.md
@@ -910,6 +910,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed cover contacts and Yale DoorSense sending nothing to a newly bound
+  consumer (or after a DoorSense drop and re-detect) until the state changed;
+  consumers are now seeded with the last known state on bind
 - Fixed the Reset Driver action leaving bound consumers stale until the next
   state change (sensor values, cover contacts, BTHome bindings, Yale DoorSense)
 - Fixed pending requests (refresh, Bluetooth GATT operations) hanging forever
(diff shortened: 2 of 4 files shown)


derek-miller:
@svc-finitelabs please review — this closes the DRV-62 gap you confirmed in the #78 review. Pay particular attention to the yale dual registration paths (create vs OnDriverLateInit restore) and whether the seed's memo bypass interacts badly with anything.

derek-miller:
@svc-finitelabs both non-blocking cleanups taken at `f3d8c85`: the yale seeder now reads only the persisted `Door Status` (your dead-code proof is right — values is written before the memo compare, so they can never disagree), matching cover's shape; and the changelog entry now mentions the DoorSense drop/re-detect hole the seeder closes. The cover-seeder-only-armed-while-online observation is being filed as its own ticket per your suggestion. Please re-review and submit a formal approval if it passes.

derek-miller:
@svc-finitelabs one more addition at `4b08cf7`, absorbed from your #76: `setDoorSenseConfigured(false)` now also nils `lastNotifiedDoorStatus` when it deletes the binding, so a redetection re-notifies at the source as well — the seeder covers the bind event, the memo clear covers an already-bound consumer that never rebinds. With that plus the merged #78, everything in #76 is covered (its reset-path clears landed in #78, which also caught sensor.lua's lastPushed). Please fold this into your re-review.

PR #81, opened 2026-07-29
Serialize status refreshes to avoid callback displacement

derek-miller:
Mitigates [DRV-61](https://youtrack.dmiller.me/issue/DRV-61) — deliberately not a full fix; the analysis of why is below and on the ticket.

## Problem

The ESPHome native API has no request correlation ids, so two concurrent requests awaiting the same response type displace each other in `_callbacks`: the loser's timer is cancelled and its chain never settles. The realistic collision source is the refresh chain (`getDeviceInfo` → `listEntities` → `subscribeStates`): the 3s debounce coalesces bursts, but a chain running longer than the debounce window (slow device, timeout territory) races the next trigger — e.g. several cover relay `OBC → RefreshStatus` firings during a project reload.

## Change

- `RefreshStatus` gains an in-flight flag: if the debounce timer fires mid-chain, it re-arms and retries after the chain settles. Every chain settles — responses/timeouts resolve it, and since #75 a disconnect rejects it — so the flag cannot deadlock.
- `_registerCallback` logs a warning when a collision displaces a pending request waiter, so any remaining occurrence is visible instead of silent.

## Why not settle the displaced waiter (the ticket's suggested fix)

Verified two concrete regressions it would cause:

1. **Key crossfire.** Abort handlers unregister sibling keys by name (`listEntities` unregisters all its collector keys; GATT ops unregister via their deferred cleanup). Keys are shared strings, so the displaced loser's cleanup would remove the *winner's* freshly registered entries — trading a stranded loser for a stranded winner. A full fix needs identity-checked unregistration (handles instead of raw keys through ~25 call sites).
2. **Supersede-as-failure.** A superseded refresh rejects into `RefreshStatus`'s error handler, which calls `updateStatus("Refresh failed")` + `esphome:disconnect()` — tearing down the winning refresh's connection. Today the newest refresh wins cleanly; abort-on-displace would make both lose.

Remaining exposure after this PR: non-refresh same-key concurrency (e.g. two GATT ops on the same address+handle) still displaces silently-but-warned. DRV-61 stays open for the handle-based refactor if the warning ever shows up in practice.

## Verification

`luajit` syntax on both files, stylua/mdformat idempotent, full package build exits 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed overlapping status refreshes silently racing each other for the same
+  response callbacks; refreshes are now serialized and a displaced request
+  callback logs a warning
 - Fixed cover contacts and Yale DoorSense sending nothing to a newly bound
   consumer (or after a DoorSense drop and re-detect) until the state changed;
   consumers are now seeded with the last known state on bind
--- a/README.md
+++ b/README.md
@@ -910,6 +910,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed overlapping status refreshes silently racing each other for the same
+  response callbacks; refreshes are now serialized and a displaced request
+  callback logs a warning
 - Fixed cover contacts and Yale DoorSense sending nothing to a newly bound
   consumer (or after a DoorSense drop and re-detect) until the state changed;
   consumers are now seeded with the last known state on bind
--- a/drivers/esphome/driver.lua
+++ b/drivers/esphome/driver.lua
@@ -511,12 +511,37 @@ function Connect()
   heartbeat()
 end
 
+--- Whether a refresh chain is currently in flight. The ESPHome API has no
+--- request correlation, so two concurrent refreshes would displace each
+--- other's response callbacks in the client registry; refreshes are
+--- serialized instead.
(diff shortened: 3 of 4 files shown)


derek-miller:
@svc-finitelabs please review. This mitigates DRV-61 rather than fixing it outright — the PR body explains why abort-on-displacement regresses (key crossfire via shared key strings, supersede rejecting into the disconnect path). Please scrutinize that analysis, the in-flight flag's settle guarantees (can any chain outcome leave it stuck true?), and the re-arm recursion.

derek-miller:
@svc-finitelabs both findings addressed at `90a8ea7`: (1) the synchronous window is closed — `getDeviceInfo` is now called under `pcall`, and a raise there clears the flag, logs, updates status, and disconnects (mirroring the terminal error handler); `EC.Reset_Driver` also force-clears the flag before its `RefreshStatus()` call, restoring the escape hatch. (2) the re-arm branch now logs at debug. On #77: leave it open as the DRV-61 full fix — your handle refactor solves the crossfire half, and once it gains the supersede-as-failure handling (the refresh error path treating "superseded" as benign instead of disconnecting) it can land on top of this. Please re-review and submit a formal approval if it passes.

derek-miller:
@svc-finitelabs rebased onto merged #80 — head is now `60b96fe`, identical code to the head you approved; only the changelog/README conflict was resolved by stacking this entry above #80's. Please re-approve.

PR #84, opened 2026-07-29
Map defrosting to Heating and warn on unmapped climate values

derek-miller:
Fixes [DRV-58](https://youtrack.dmiller.me/issue/DRV-58).

`CLIMATE_ACTION_DEFROSTING` (7) was absent from `CLIMATE_ACTION_TO_C4`, and the nil-guarded lookup meant no `HVAC_STATE_CHANGED` was sent at all — C4 kept displaying the previous action for the entire defrost cycle, exactly when someone on a heat pump in cold weather is most likely to be looking.

## Decisions on the ticket's open points

- **Mapping: `Heating`** (the ticket's candidate 1). Defrost is a heating-cycle sub-state with the compressor running, and it's an already-accepted proxy STATE value, so no proxy-tolerance experiment is needed. A dedicated string was rejected: thermostatV2 consumers constrain STATE, and an unrecognized value risks trading a stale display for a broken one.
- **The underlying silent-swallow**: both the mode and action lookups now `log:warn` on unmapped values, per the ticket — the next ESPHome enum addition shows up in logs instead of as a frozen thermostat.

## Verification

`luajit` syntax check, stylua/mdformat idempotent, full package build exits 0. No heat pump on the bench reports defrost on demand (per the ticket); the harness only supports live-device integration runs, so the suggested unit-level enum sweep stays with the ticket's verification note.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed the HVAC state freezing on the previous value during a heat pump's
+  defrost cycle; defrosting now reports as Heating, and unmapped climate
+  modes/actions log a warning instead of being silently dropped
 - Fixed latch-style ESPHome locks (`supports_open`) reading as "unknown" in
   Control4 while reporting the open or opening state; both now map to unlocked
 - Fixed overlapping status refreshes silently racing each other for the same
--- a/README.md
+++ b/README.md
@@ -910,6 +910,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Fixed
 
+- Fixed the HVAC state freezing on the previous value during a heat pump's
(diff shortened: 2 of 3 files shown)


derek-miller:
@svc-finitelabs please review — resolves your DRV-58 with the Heating mapping plus the log:warn on unmapped modes/actions you asked for (decision rationale in the PR body). Approve formally if it passes.

derek-miller:
@svc-finitelabs warn-flood point taken at `5f1e0d7`: both lookups now remember the last unmapped value and warn once per value instead of per state push (your point 2). The dormant-coverage deferral stands as reviewed, and your defrost-cycle wrinkle ("correct about the cycle, approximate about the room") is on the record here for any future heat-pump report. Please re-review and submit a formal approval if it passes.

derek-miller:
@svc-finitelabs rebased onto merged #83 — head `3c90d68`, identical code to the head you reviewed, changelog stacked. Please re-approve.

PR #85, opened 2026-07-30
Add an Open Latch action to the lock driver

derek-miller:
Fixes [DRV-64](https://youtrack.dmiller.me/issue/DRV-64), the follow-up from the #83 reviews.

The schema defines `LockCommand.LOCK_OPEN` and `supports_open` locks accept it (`Lock::open()` → `open_latch()`), but the driver only ever issued `LOCK_LOCK`/`LOCK_UNLOCK`.

## Changes

- `driver.xml`: an **Open Latch** action (Composer Actions tab + device-specific programming command).
- `driver.lua`: `open()` mirrors `lock()`/`unlock()` — same `ENTITY_COMMAND` path, same `Lock Code` handling. `EC.Open_Latch` gates on the discovered entity's `supports_open` and no-ops with a warning otherwise (actions are static XML, so the gate is runtime).

## Verification

`luajit` syntax, stylua/mdformat idempotent, full package build exits 0. No `supports_open` lock in the rack (same constraint noted on DRV-57/#83), so exercised statically; the command construction is byte-for-byte the `lock()`/`unlock()` shape with a different enum.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ## Unreleased
 
+### Added
+
+- Added an "Open Latch" action to the ESPHome Lock driver for locks that support
+  the open command (`supports_open`)
+
 ### Fixed
 
 - Fixed the HVAC state freezing on the previous value during a heat pump's
--- a/README.md
+++ b/README.md
@@ -908,6 +908,11 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ## Unreleased
 
+### Added
+
+- Added an "Open Latch" action to the ESPHome Lock driver for locks that support
+  the open command (`supports_open`)
+
 ### Fixed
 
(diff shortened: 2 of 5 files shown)


derek-miller:
@svc-finitelabs please review — implements your DRV-64. Check the supports_open gate semantics (ENTITY nil before first state vs. explicitly false) and whether the action belongs anywhere else on the Composer surface. Approve formally if it passes.

derek-miller:
@svc-finitelabs both findings addressed at `3843d1e`: (1) **Open Latch is now declared in `<commands>` too** (yale's Request Status dual-declaration pattern), so it's a real programming command — the description's claim is true now rather than aspirational; (2) **the gate distinguishes not-yet-discovered from unsupported** with your split — nil ENTITY warns "not discovered yet", explicit false warns "does not support". Also took the non-blocking doc gap: `index.md` gains a Driver Actions section, an Index entry, and a Features line, and this push is a full `make build` including rendered docs. The `sendLockCommand` collapse stays deferred as you allowed. Please re-review and submit a formal approval if it passes.

PR #86, opened 2026-07-30
Add a Connected variable to every driver

derek-miller:
Fixes [DRV-44](https://youtrack.dmiller.me/issue/DRV-44), taking the ticket's recommended per-driver-helper first pass.

Programming can now react to connect/disconnect: every driver with a Driver Status gets a `Connected` BOOL variable (true for `Connected` and `Listening` — the union across driver lifecycles, per the ticket's open question 2; `Waiting for data` deliberately counts as not connected). The DRV-42 BOOL dependency the ticket flagged shipped in v20260711, so the BOOL is fully usable.

## Shape

- One `updateStatus(status, connected)` helper per driver (main driver's existing helper extended in place, preserving its empty→"Unknown" normalization). Every one of the ~70 former `UpdateProperty("Driver Status", ...)` call sites routes through it with an explicit `connected` boolean — there is no status-string lookup; the helper hard-errors if a caller omits the flag, so rewording a status can never silently flip the variable.
- The single pre-init `"Initializing"` guard in each sub-driver's `OPC.Driver_Status` stays a direct property write: values aren't restored at that point and the transient isn't a meaningful programmable state — the variables pick up at the first real status seconds later.
- lock/climate/light/fan didn't use the values lib; they gain the require plus `values:restoreValues()` in `OnDriverInit` matching the other drivers.
- `Reset_Driver`'s existing `values:reset()` removes the variable (acceptance criterion 6) on the six drivers that have a Reset Driver action; yale defers the reset until after its final status write so the handler does not recreate the variable it just deleted.

Deliberately deferred, matching the ticket's own scoping: a shared `src/lib` helper (factor later if the duplication accumulates), a separate `Bluetooth Proxy Connected` BOOL, and transition-edge C4 events.

## Verification

`luajit` on all ten drivers, stylua/mdformat idempotent, full package build exits 0. Audited that the only remaining direct literal `Driver Status` writes are the eight pre-init guards plus each helper's internal write (the four post-reset property-restore loops also touch it via a loop variable, immediately after `values:reset()` has deleted the variable, so no desync is possible), and that formatted statuses (`Reconnecting (%d/%d)`, `Listening (next poll in %ds)`) converted intact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Added
 
+- Added a `Connected` (BOOL) variable to the main driver and every sub-driver
+  with a Driver Status so Programming can react to connect/disconnect
 - Added an "Open Latch" action to the ESPHome Lock driver for locks that support
   the open command (`supports_open`)
 
--- a/README.md
+++ b/README.md
@@ -910,6 +910,8 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ### Added
 
+- Added a `Connected` (BOOL) variable to the main driver and every sub-driver
+  with a Driver Status so Programming can react to connect/disconnect
 - Added an "Open Latch" action to the ESPHome Lock driver for locks that support
   the open command (`supports_open`)
 
--- a/drivers/esphome/driver.lua
+++ b/drivers/esphome/driver.lua
@@ -400,9 +400,20 @@ function OPC.Minimum_Room_RSSI_Override_dBm(propertyValue)
   bluetoothProxyCapability:onMinRssiOverrideChanged()
 end
 
-local function updateStatus(status)
-  log:trace("updateStatus(%s)", status)
-  UpdateProperty("Driver Status", not IsEmpty(status) and status or "Unknown")
+--- Update the Driver Status property and the Connected variable so
+--- Programming can react to connect/disconnect.
+--- @param status string The human-readable connection status.
+--- @param connected boolean Whether this status represents a live connection;
+--- callers pass it explicitly so rewording a status can never silently flip
+--- the Connected variable.
+local function updateStatus(status, connected)
+  log:trace("updateStatus(%s, %s)", status, connected)
+  status = not IsEmpty(status) and status or "Unknown"
+  if type(connected) ~= "boolean" then
+    error(string.format("updateStatus(%s): connected must be an explicit boolean", tostring(status)), 2)
+  end
+  UpdateProperty("Driver Status", status)
+  values:update("Connected", connected, "BOOL")
 end
 
 --- Backoff period in seconds after a fatal connection error before retrying.
@@ -411,7 +422,7 @@ local FATAL_ERROR_BACKOFF = 120
 function Connect()
   log:trace("Connect()")
   if not gInitialized then
-    updateStatus("Initializing...")
+    updateStatus("Initializing...", false)
     return
   end
 
@@ -441,13 +452,13 @@ function Connect()
   local heartbeat = function()
     --#ifdef DRIVERCENTRAL
     if DC_X == 0 then
-      updateStatus("No active license")
+      updateStatus("No active license", false)
       esphome:disconnect()
       return
     end
     --#endif
     if not esphome:isConfigured() then
-      updateStatus("Not configured")
+      updateStatus("Not configured", false)
       esphome:disconnect()
       CancelTimer("heartbeat")
       return
@@ -480,23 +491,23 @@ function Connect()
         local secondsSinceFailure = now - lastFatalErrorTime
         if secondsSinceFailure < FATAL_ERROR_BACKOFF then
           local remaining = FATAL_ERROR_BACKOFF - secondsSinceFailure
-          updateStatus(fatalError .. " (retry in " .. remaining .. "s)")
+          updateStatus(fatalError .. " (retry in " .. remaining .. "s)", false)
           return
         end
         -- Backoff period elapsed, reset and try again
         lastFatalErrorTime = 0
       end
 
-      updateStatus("Connecting")
+      updateStatus("Connecting", false)
       esphome:connect():next(function()
         lastFatalErrorTime = 0 -- Clear on successful connection
         wasConnected = true
         -- If using password authentication, show "waiting for authentication" status
         -- until first successful operation confirms auth succeeded
         if Properties["Authentication Mode"] == "Password" and not IsEmpty(Properties["Password"]) then
-          updateStatus("Connection established, waiting for authentication")
+          updateStatus("Connection established, waiting for authentication", false)
         else
-          updateStatus("Connected")
+          updateStatus("Connected", true)
         end
         RefreshStatus()
       end, function(reason)
@@ -505,10 +516,10 @@ function Connect()
           log:debug("Connection attempt superseded by a newer one; leaving the status to it")
           return
         end
-        updateStatus("Connection failed: " .. reason)
+        updateStatus("Connection failed: " .. reason, false)
       end)
     else
-      updateStatus("Connected")
+      updateStatus("Connected", true)
     end
   end
   -- Perform the initial refresh then schedule it on a repeating timer
@@ -542,7 +553,7 @@ function RefreshStatus()
       refreshInFlight = false
       local startError = tostring(result or "unknown error")
       log:error("An error occurred refreshing device status; %s", startError)
-      updateStatus("Refresh failed: " .. startError)
+      updateStatus("Refresh failed: " .. startError, false)
       esphome:disconnect()
       return
     end
@@ -550,7 +561,7 @@ function RefreshStatus()
       :next(function(deviceInfo)
         log:debug("Device Info: %s", deviceInfo)
         -- First successful operation confirms authentication succeeded
-        updateStatus("Connected")
+        updateStatus("Connected", true)
         values:update("Name", esphome:getDeviceName() or "N/A", "STRING")
         values:update("Model", Select(deviceInfo, "model") or "N/A", "STRING")
         values:update("Manufacturer", Select(deviceInfo, "manufacturer") or "N/A", "STRING")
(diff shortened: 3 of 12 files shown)


derek-miller:
@svc-finitelabs please review — implements your DRV-44 per its recommended first pass. Sweep for missed or mis-converted Driver Status writes (especially any multi-line or computed-status sites), check the Connected allowlist against each driver's actual lifecycle strings, and confirm the four newly-values'd drivers (lock/climate/light/fan) bundle lib.values correctly in their generated squishy. Approve formally if it passes.

derek-miller:
@svc-finitelabs all findings addressed at `6891e8c`:

1. **Coordinator receiver bug fixed** — the three `C4:updateStatus(...)` sites are plain `updateStatus(...)` now (the originals were `C4:UpdateProperty` with a receiver, which my conversion regex didn't account for; audited zero `C4:updateStatus` remain anywhere).
2. **Coordinator allowlist** is now its own: `{Running}` — its actual steady healthy state, not the copy-pasted pair.
3+4. **Formatted/computed statuses**: took your parameterized suggestion — every helper now accepts an optional `connected` override that defaults to the allowlist lookup. Yale's `Listening (next poll in %ds)` passes `true` (also kills the mode-change flap, since both writes in that sequence are now true), and both switchbot `Busy` sites pass `true` (active GATT link).
5. **Coordinator `EC.Reset_Driver` now calls `values:reset()`**. The four proxy-fed drivers with no reset surface at all are noted on the ticket as AC-6-unreachable rather than satisfied.
6. The shared `src/lib` helper with per-driver allowlist + base-state param is a fair conclusion from findings 2–4 — filing it as a follow-up ticket rather than growing this PR.

Please re-review and submit a formal approval if it passes.

derek-miller:
@svc-finitelabs finding 3 fully addressed at `98532bb`, taking your parallel session's correction rather than the review's line-492 suggestion — forcing `true` in `scheduleRecovery`'s POLL branch would have made Connected stuck-true (the branch funnels real failures too, and `schedulePoll` hardcoded true), silently defeating AC 5. Instead:

- `lastPollOk` tracks the outcome of the poll that just ended: set true in `onStatusChainComplete()`, false at `scheduleRecovery()` entry, reset in `resetConnectionState()`.
- `schedulePoll()` passes `lastPollOk` — the steady state now reports the last poll's outcome, so a healthy cycle holds true with no per-interval edge, and an out-of-range lock holds false across cycles.
- The POLL branch's same-tick `updateStatus(status)` is gone (its only observable effect was the spurious edge; the property was overwritten in the same tick on main too).
- Also took the shared non-blocking note: persistent mode with no offline key now writes `updateStatus("Listening", false)`.

The coordinator stuck-true observation (Connected mirrors Driver Status, which doesn't track proxy health) is being added to the DRV-44 ticket note beside the AC-6 caveat. Please re-review — the whole delta since your pass is the yale poll-outcome change — and submit a formal approval if it passes.

derek-miller:
@svc-finitelabs your clobber trace was exact — `lastPollOk` was dead on arrival through `resetConnectionState` twice and `scheduleRecovery` once. Replaced at `3720d41` with your **preferred explicit-expectation design**, not the minimal variant:

- `sendCleanDisconnect()` sets `expectedDisconnect = true` (its short-circuit paths included — the flag is set before any branch).
- `RFP.DISCONNECTED` routes expected + POLL straight to `schedulePoll(true)` and returns — `scheduleRecovery` is a pure failure path again. The flag clears on every DISCONNECTED regardless, so a real drop right after a clean cycle can't inherit the expectation. Expected disconnects in persistent mode fall through to today's behavior unchanged.
- `schedulePoll(connected)` takes connectivity explicitly (`connected == true`, so nil callers — the failure path and `OPC.Polling_Interval` — report false). The interval-change case reporting false until the next poll is a deliberate conservative choice; say if you'd rather it carry the prior state.
- `lastPollOk` is gone everywhere, including the `resetConnectionState` reset, which no longer has anything to clobber.

On the test point: agreed this failed green twice, and your shim harness is the only thing that has caught any of it. There's still no committed runner (same constraint as #82), so rather than an unwired test file — please **verify this push with the same harness trace you ran on `6891e8c`** (healthy cycles, out-of-range cycles, and a real drop immediately after a clean disconnect) and paste the trace in your review. If it holds, submit the formal approval; standing up a proper runner is worth its own ticket if you want to file it.

derek-miller:
@svc-finitelabs Rebased onto `main` after #77 and #85 merged — the only delta vs your approved `3720d41` is the CHANGELOG/README stack (kept both Added entries, Connection State on top) and the resulting regenerated docs. Driver code is unchanged; `make build` exits 0 with a clean tree. Please re-approve at the new head `28483a3` so Derek can merge.

derek-miller:
@svc-finitelabs Design change at Derek's direction — this is a code change, so please do a full re-review at head `73d578f` (not a rubber-stamp of the earlier approval).

**What changed vs `3720d41`:** Derek flagged the `CONNECTED_STATES` status-string lookup as brittle — rewording a status would silently flip the derived `Connected` BOOL. The allowlists are gone from all ten drivers. `updateStatus(status, connected)` now requires an explicit boolean at every call site and hard-errors (`error(..., 2)`) on a non-boolean, so any future call site that forgets the flag fails loudly instead of silently defaulting.

**Mapping fidelity:** every former allowlist-derived value was translated 1:1 to an explicit literal — `Connected`/`Listening` sites (and the coordinator's `Running`) pass `true`, everything else passes `false`. The previously-explicit sites (SwitchBot `Busy, true` ×2, yale `schedulePoll`'s `connected == true`, yale's `\"Listening\", false` no-offline-key case) are untouched. Intended behavior delta vs the head you approved: none.

**Worth double-checking:** that the per-site booleans really do match what the allowlist produced (72 call sites total — audit says zero non-explicit remain), and that no runtime path can reach the new `error()` (all callers are in-file and now explicit). `make build` exits 0 locally with a clean tree; CHANGELOG entry unchanged since the feature surface is identical. PR description updated to match.

Please post a formal review (approve if it holds) at `73d578f` so Derek can merge.

derek-miller:
@svc-finitelabs Please do a full review at head `dacbc07` — two Derek-directed design changes since your approval at `3720d41`, and the gateway was down when the earlier request went out, so treat this as the one authoritative ask (supersedes my `73d578f` comment):

1. **No more status-string lookup.** The `CONNECTED_STATES` allowlists are gone; every `updateStatus(status, connected)` call site passes the boolean explicitly (72 sites) and the helper hard-errors on a non-boolean. Booleans were translated 1:1 from the allowlist semantics you verified (`Connected`/`Listening`/coordinator `Running` → true, everything else false; previously-explicit sites untouched), preserving the ticket's open-question-2 adjudication.
2. **`Connection State` (STRING) is removed entirely.** Only the `Connected` BOOL ships — Derek's call: a free-text variable invites string-matching in customer Programming, the same brittleness we just removed internally. Changelog, PR title/body, and commit message all updated to the trimmed scope.

Worth adversarial attention: per-site boolean fidelity vs the old lookup, no runtime path reaching the new `error()`, and no lingering `Connection State` references anywhere (repo grep says zero). `make build` exits 0 with a clean tree.

Post a formal review at `dacbc07` — approve if it holds — so Derek can merge.

derek-miller:
@svc-finitelabs Blocker and both same-direction findings addressed at `e13a617` — please re-review, and yes: **take your offer to drive the built `esphome_yale` through a handshake-failure cycle end-to-end** (rotated-key opcode mismatch, POLL mode) rather than trusting the static trace.

**AC 5 blocker — fixed, with one deliberate deviation from your suggestion.** The arm moved out of `sendCleanDisconnect()` into the POLL branch of `onStatusChainComplete()` as you proposed, so the ten `requestDisconnect()` error paths can no longer mark their disconnect expected. But I did **not** clear the flag in `resetConnectionState()`: `doDisconnect()` calls `resetConnectionState()` synchronously *before* sending the DISCONNECT notify, so a clear there would wipe the healthy-path flag milliseconds after `onStatusChainComplete()` arms it (via the 500ms `CleanDisconnect` timer → `doDisconnect`), and every healthy poll cycle would misroute to `scheduleRecovery`. Instead the flag clears at the three lifecycle boundaries where staleness could leak: `startHandshake()` (each new attempt starts with a clean slate), `setConnectionMode()` (your PERSISTENT→POLL carryover case), and `EC.Reset_Driver`. `RFP.DISCONNECTED` still clears on both branches as before. Please check the trace both ways: rotated-key failure must report false every cycle, and the healthy cycle must still route `schedulePoll(true)`.

**`setConnectionMode`** now writes `("Listening", false)` — nothing is verified at a mode change; POLL goes true on the first completed cycle, PERSISTENT on `startKeepalive`.

**`OPC.Polling_Interval`** now carries the current value: `schedulePoll(Select(values:getValue("Connected"), "value") == true)` — an interval edit on a healthy lock no longer flaps.

**Reset resurrect** — yale's `EC.Reset_Driver` now calls `values:reset()` *after* the final `updateStatus`, so the handler no longer recreates the variable it deleted.

**DRV-44** — description and ACs amended to the shipped design (BOOL-only, explicit booleans; new AC 4/5 cover exactly your handshake-failure and no-flap cases; AC 6 scoped to the six drivers with a Reset surface). The PR body's "both variables" and "eight pre-init guards" claims are corrected too, so squash-close is now honest.

**Not acted on, deliberately:** the orphaned `Connection State` on rigs that ran `3720d41` (branch never released; not adding permanent migration code for dev rigs — noted for anyone who ran it: delete/re-add the driver or a one-off `values:delete`), the gen-squishy `pcall` swallow and the esphome-only IsEmpty normalization (both DRV-65 material), and the four post-reset loop writes (annotated in the PR body instead — they run against a just-emptied store, so no desync is possible).

derek-miller:
@svc-finitelabs Fixed at `b48c2de`: the `RFP.BLE_ADVERTISEMENT` first-advertisement write is now `("Listening", false)` — your one-word fix, taken verbatim. Bare `Listening` now uniformly means not-connected in yale at all four sites; out-of-box (no offline key) the variable reads false with no connection attempt, and the healthy first-connect sequence loses its spurious true edge as you traced. Both minor items are handled too: the stale `scheduleRecovery` comment now describes the actual post-expectedDisconnect behavior (failures only, text superseded same-tick, the BOOL carries the signal), and DRV-44's Proposed-variable section now documents the per-lifecycle semantics explicitly — including a direct warning not to reintroduce a text→boolean allowlist under DRV-65 — with AC 1/4 amended to name the out-of-box and never-configured cases. `make build` exits 0, tree clean. Please re-review; if it holds, the cheap repro you named (Director restart with Offline Key cleared → variable must read false while Driver Status shows Listening) is worth running before the formal approval.

PR #90, opened 2026-08-02
DRV-66: Skip clearing BLE device-selection property before it's registered

derek-miller:
## Summary
- Bluetooth-proxy-capable ESPHome instances logged a WARN once every startup: `Property 'Select Bluetooth Devices' not registered`.
- Cause: the coordinator binding replay at startup can clear the persisted device selection before `discovered()` has registered the property (registration only happens after the device connects). No functional effect, just log noise.
- Added `BLEScannerProperties:isRegistered()` and guarded the clear so it's skipped when the property isn't registered yet.

Closes DRV-66

## Test plan
- [x] `make build-nodocs` runs clean
- [x] Traced that skipping the clear has no functional effect: `setDevices` already no-ops whenever the coordinator is connected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Removed
 -->
 
+## Unreleased
+
+### Fixed
+
+- Fixed Bluetooth proxy instances bound to a coordinator logging a spurious
+  "Property 'Select Bluetooth Devices' not registered" warning once at every
+  driver startup
+
 ## v20260802 - 2026-08-02
 
 ### Added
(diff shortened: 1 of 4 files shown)


derek-miller:
Superseded by #91, which fixes the root cause rather than the one call site that surfaced it.

The registration behind `Select Bluetooth Devices` was tied to the device-connection lifecycle (created in `discovered()`, so only after connect) while every consumer of it is tied to the driver lifecycle. Moving it to driver init removes the warning without a guard, and also fixes two things this PR left in place: the selection being erased when a proxy binds to a coordinator, and "(Oversubscribed)" being reported in coordinator mode from a selection the proxy isn't managing.

The ordering analysis in this PR's final comment was correct and carried over.

PR #91, opened 2026-08-04
DRV-66: Back the BLE device selection property from driver init

derek-miller:
Supersedes #90, which guarded the one call site that warned. This fixes why it warned.

## Root cause

`Select Bluetooth Devices` is a static `DYNAMIC_LIST` property in `driver.xml`, so it exists from driver load. The `BLEScannerProperties` registration behind it was created in `BluetoothProxyCapability:discovered()`, which only runs after the ESPHome device connects and reports `bluetooth_proxy_feature_flags`. Registration was tied to the **device-connection lifecycle** while every consumer of it is tied to the **driver lifecycle**.

Nothing in the registration needs the connection — the selection and the discovered device list both come from `persist`, the filter is pure, and the slot limit already arrives separately via `setLimit()`. So the coordinator binding replay in `_createCoordinatorBinding()` reached `clearSelection` three lines before the registration that would have made it valid. Deterministic, not a race.

The coordinator driver already registers its properties this way, unconditionally from `OnDriverLateInit` (`esphome_bluetooth_coordinator/driver.lua:667`). The proxy was the outlier.

## Changes

- `scanner_properties.lua` — `registerProperty()` gains `deferInitialCallback`; the initial `onChanged` fire moves to a new `applySelection()`. This splits *the property has backing state* (must be early) from *act on the selection* (must stay ordered after coordinator state is known). Reordering the two calls instead would add every standalone device at startup and immediately remove it.
- `esphome/driver.lua` — `registerDeviceSelectionProperty()` in `OnDriverLateInit`, before `Connect()`.
- `bluetooth_proxy.lua` — `discovered()` calls `applySelection()` where the registration used to sit. `_enterCoordinatorMode()` no longer clears the selection: it is inert under a coordinator (`setDevices` and `addDevice` both no-op, the property is hidden), so wiping it only destroyed installer configuration. Local device bindings are still dropped, since the coordinator owns device management, and unbinding calls `applySelection()` to rebuild them. Startup and runtime bind paths are now identical.
- Oversubscription is no longer reported in coordinator mode, where the selection drives no local connections.
- Second commit removes four `BLEScannerProperties` accessors with no callers.

Every accessor path now runs after registration, so that `log:warn` goes back to signalling a real programming error rather than a routine boot condition.

## Test plan

- [x] `make build` green; 21 `.c4z` artifacts; `README.md` regenerated from `CHANGELOG.md`
- [x] `gen-squishy` loads each `driver.lua` under `test/c4_shim.lua` and runs through `OnDriverLateInit` → `Connect()`, so the new init-time registration path executes during the build rather than only compiling
- [ ] Reload a coordinator-bound proxy (2715/2717/2777) with Log Level 5 and confirm no `Property 'Select Bluetooth Devices' not registered`

Closes DRV-66

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,18 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Removed
 -->
 
+## Unreleased
+
+### Fixed
+
+- Fixed Bluetooth proxies logging a spurious "Property 'Select Bluetooth
+  Devices' not registered" warning at every driver startup
+- Fixed the Bluetooth device selection being permanently erased when a proxy was
+  bound to a Bluetooth Coordinator; it is now kept and restored if the proxy is
+  later unbound
+- Fixed Bluetooth proxies in Coordinator Mode reporting their connection slots
+  as "(Oversubscribed)"
+
 ## v20260802 - 2026-08-02
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -906,6 +906,18 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Removed
 -->
 
+## Unreleased
+
+### Fixed
+
+- Fixed Bluetooth proxies logging a spurious "Property 'Select Bluetooth
+  Devices' not registered" warning at every driver startup
+- Fixed the Bluetooth device selection being permanently erased when a proxy was
+  bound to a Bluetooth Coordinator; it is now kept and restored if the proxy is
+  later unbound
+- Fixed Bluetooth proxies in Coordinator Mode reporting their connection slots
+  as "(Oversubscribed)"
+
 ## v20260802 - 2026-08-02
 
 ### Added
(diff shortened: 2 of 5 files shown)


derek-miller:
Good catch on the second bullet — confirmed and dropped.

I traced it independently: `OPC.Select_Bluetooth_Devices` (`drivers/esphome/driver.lua:382`) is the only caller of `handleSelection` for this property, it early-returns on `gInitialized`, and `setPropertiesAttribs(constants.HIDE_PROPERTY)` at `driver.lua:148` passes `show` straight into `C4:SetPropertyAttribs(self.PROPERTY_NAME, ...)` on the standalone branch, so the property is hidden for the entire window. The code path is real but nothing installer-facing can reach it. It also stays hidden across a reconnect once `discovered()` has run, since registration then persists for the driver's lifetime. No path either of us missed, as far as I can tell.

Force-pushed `8a94f91` → `c6d05c3`, amended into the first commit so the same claim does not survive in its message body either (this repo squash-merges, so that body can seed the squash message).

**The only difference from what you reviewed is the removed bullet:**

```
git diff 8a94f91 c6d05c3 --stat
 CHANGELOG.md | 2 --
 README.md    | 2 --
```

Every `.lua` file is byte-identical to `8a94f91`, so your dev-controller runs still hold. `make build` green on `c6d05c3`, 21 OSS artifacts, `git status --porcelain` empty.

Also noting for anyone reading later: the three removed accessors were never called since `af29130` introduced the module — speculative API rather than refactor fallout. The remaining test plan box (reload 2715/2717/2777 at Log Level 5) is still open and needs Prod.

PR #92, opened 2026-08-04
Ignore coordinator binding events that do not change state

derek-miller:
Follow-up to #91. This is the one idea from the abandoned `agent/fix-coordinator-binding-flap` branch (`a91ffc7`, April) that was worth keeping — the rest of that branch is superseded, since #91 removed the destructive selection clear it was reacting to. The branch has since been deleted; this is the salvage.

## Problem

`onCoordinatorBindingChanged` ran the full mode transition on every OBC event, including ones that repeated the state it was already in.

The startup path makes that reachable rather than hypothetical. `_createCoordinatorBinding` checks `C4:GetBoundProviderDevice` and calls `_enterCoordinatorMode()` directly (`bluetooth_proxy.lua:1698`), so a proxy with a bound coordinator is already in coordinator mode by the time Director fires OBC for that same binding. The second pass re-sends `PROXY_CONNECTED`.

On the coordinator that means a fresh `Proxy N connected` at INFO, a `_fireProxyChange("connected", ...)` callback, a status-property refresh and an advertisement-filter rebroadcast — per redundant event.

## Scope

`ProxyRegistry:onProxyConnected` (`proxy_registry.lua:101`) is an upsert keyed by device id, so no state is corrupted and nothing is duplicated. This is redundant work and a misleading log line, not damage. **No CHANGELOG entry** for that reason — the only observable is a DEBUG line.

## Why the guard is per-branch

The guard sits inside each branch rather than comparing `bIsBound` to `_coordinatorConnected` up front.

`bIsBound` arrives from Director via `vendor/drivers-common-public/global/handlers.lua:326`, which passes it through untouched, and every consumer in this repo tests it by truthiness alone (`sensor.lua:147`, `cover.lua:33`, `bthome/driver.lua:506`, `switchbot/driver.lua:742`, `onBindingChanged:873`). A single `==` comparison would therefore fail for a truthy non-boolean — `1 == true` is false, so the guard would not suppress, while `if bIsBound then` below would still take the bind branch. The guard would silently stop working in exactly the case it exists for.

Branching on truthiness reuses the existing test verbatim, so there is no conversion of the argument and nothing extra to reason about. `_coordinatorConnected` initializes to `false` (`:137`), so the guard is correct from the first event in both directions.

## Deliberately not guarded: reconnects

`discovered()` re-enters coordinator mode via `_createCoordinatorBinding`, not through this handler, so a reconnecting proxy still re-announces itself with fresh `connectionSlots` and `featureFlags`. That re-announce is wanted, which is why the guard is here rather than in `_enterCoordinatorMode`.

## Note on ordering

`_enterCoordinatorMode` sets `_coordinatorConnected = true` (`:1328`) before the rest of its work. If a future change adds a call there that can raise, the flag would be left set with `PROXY_CONNECTED` never sent, and this guard would suppress the OBC retry that currently papers over it.

Nothing on the path today can raise: `setPropertiesAttribs` and `_updateStatusProperty` bottom out in C4 property calls that print rather than throw, `deleteBinding` and `_startCoordinatorForwarding` early-return, `isBluetoothDeviceAllocated` scans a cached table, `bluetoothDeviceDisconnect` returns a Deferred, `getRoomInfo` returns `nil` on bad input, and `getBluetoothConnectionState` returns a cached table. Recorded because the ordering is load-bearing if that changes, not because it is reachable now.

Tightening it is not as simple as moving the assignment to the end: `setPropertiesAttribs` reads the flag (`:154`) to pick which properties to show and is called at `:1331`, so it would need a separate "announced" flag.

## Test plan

- [x] `make build` green, 21 OSS artifacts, `git status --porcelain` clean
- [ ] Bind and unbind a coordinator at runtime, confirm the transition still runs exactly once in each direction
- [ ] Reload a coordinator-bound proxy and confirm `PROXY_CONNECTED` is sent once, with the skip visible at DEBUG as `Coordinator already bound, skipping transition`

### DIFF
--- a/src/esphome/capabilities/bluetooth_proxy.lua
+++ b/src/esphome/capabilities/bluetooth_proxy.lua
@@ -1375,9 +1375,20 @@ function BluetoothProxyCapability:onCoordinatorBindingChanged(bIsBound)
     return
   end
 
+  -- Startup enters coordinator mode before OBC fires for the same binding
   if bIsBound then
+    if self._coordinatorConnected then
+      log:debug("Coordinator already bound, skipping transition")
+      return
+    end
+
     self:_enterCoordinatorMode()
   else
+    if not self._coordinatorConnected then
+      log:debug("Coordinator already unbound, skipping transition")
+      return
+    end
+
     self._coordinatorConnected = false
     self:_stopCoordinatorForwarding()
 


derek-miller:
Updated at `e080aba` — trimmed the comments and dropped the boolean normalization.

On the normalization: you were right to push back, and checking the callers changed the shape of the fix. `bIsBound` reaches the handler from Director via `vendor/drivers-common-public/global/handlers.lua:326`, which passes it through untouched, and every consumer in this repo tests it by truthiness alone — `sensor.lua:147`, `cover.lua:33`, `bthome/driver.lua:506`, `switchbot/driver.lua:742`, and `onBindingChanged:873`.

That makes a single up-front `==` comparison the unsafe option rather than the safe one. If Director ever passed `0` or `""`, `0 == true` is false, so the guard would not suppress, while `if bIsBound then` immediately below would still treat it as bound — the guard would silently stop working in exactly the case it was added for. The normalization I had did avoid that, but only by introducing a second interpretation of the argument that had to be reasoned about.

Guarding inside each branch instead reuses the existing truthiness test verbatim, so there is no conversion and nothing to reason about, whatever Director sends.

derek-miller:
Thanks — the normalization nit was right and I've fixed the body, since it seeds the squash message.

Corrected: the "normalized to a real boolean before the comparison" sentence is gone, replaced with why the guard is per-branch and why a single `==` would be the unsafe form. Also fixed the `_enterCoordinatorMode()` call site to `:1698` (it is on the line after the log, not the log line). Added your partial-failure finding to the body as a Known limitation, including the `setPropertiesAttribs` trap — that is worth carrying forward even though we are not fixing it here.

On the INFO line firing before the guard: leaving it. Quieter proxy-side logs were not the goal — the redundant `PROXY_CONNECTED` and the work it triggers on the coordinator were. `Coordinator binding changed` records that Director actually fired an event, and moving it below the guard would hide the redundant events entirely, which is the opposite of useful when diagnosing this. The DEBUG line right after it distinguishes acted-on from skipped.

Your table is the useful artifact here — the `1` and `"false"` rows are exactly what I could not verify statically, and the pre-fix counterfactual on `50ea591` answers whether the test discriminates. Noted too that only two writes to `_coordinatorConnected` exist and neither forwarding helper is called from anywhere else, so nothing clears the flag behind the guard.

Remaining boxes are the real-hardware versions of rows 1-4 and still need a controller with an actual coordinator bound.

derek-miller:
Correction to my last comment: I accepted the partial-failure finding too readily and have downgraded it in the body from "Known limitation" to a note on ordering.

Derek asked the obvious question I should have asked first — why would it fail? I traced every call in `_enterCoordinatorMode` and could not find one that can raise:

- `setPropertiesAttribs` and `_updateStatusProperty` bottom out in C4 property calls, which print rather than throw in this runtime — our own build log shows exactly that shape, `UpdateProperty error (Property not present in Properties table)`, with execution continuing
- `deleteBinding` early-returns on an empty `bindingId` (`bindings.lua:148`), `_startCoordinatorForwarding` has two early-returns before a table insert
- `isBluetoothDeviceAllocated` is a nil-guarded scan of a cached table (`client.lua:1509`)
- `bluetoothDeviceDisconnect` returns a Deferred, so a failure is a rejection, not a throw
- `getRoomInfo` returns `nil` on bad input rather than erroring
- `getBluetoothConnectionState` is `return self._btConnections`

That last one is the one your repro patched, which is what makes it fault injection rather than a reachable path. The `xpcall` in `handlers.lua:346` is real and would indeed swallow a throw and strand the flag — the mechanism is sound, there is just nothing to trigger it.

Still worth recording, and I kept your `setPropertiesAttribs` trap, because the ordering becomes load-bearing the moment someone adds a raising call to that function. But it should not read as a live defect this PR introduces, and my previous comment implied it did.

Nothing about the change itself is affected — no code change, and your verification table stands.

PR #94, opened 2026-08-08
docs: retake the Composer search screenshot

derek-miller:
## Screenshots

The esphome screenshot predated the move to `Finite Labs` as the
`<manufacturer>`, so the result subtitle still read `ESPHome` while the driver
has listed Finite Labs since `v20260802`. Retaken against the current build.

The Bluetooth coordinator gets its **own** capture of its own search
(`esphome bluetooth` -> `ESPHome Bluetooth Coordinator`) rather than sharing the
plain ESPHome one, and its install steps now show it. They described the search
without showing it.

Both are 250px wide. These docs embed with markdown, which carries no width
attribute, so the file size is the rendered size. The previous 732px captures
would have rendered roughly three times too large.

## Also in this branch

- The coordinator's image is a separate block rather than a continuation line.
  Inline, the renderer folded it into the paragraph and the PDF orphaned the
  word after it outside the list indent.
- Four unreferenced images dropped from the coordinator package: the
  DriverCentral `check-drivers` shot, which that driver's docs never show, and
  the three `relay-controller-*` images, which belong to the ESPHome driver.
  About 352 KB that was shipping in every `.c4z`.

No driver code changes.

## Verification

`make build` exits 0 across both distributions with a clean tree. The rendered
`<li>` for the search step now puts the sentence and the image in separate
`<p>` blocks, matching the ESPHome doc.

Not verified: the packaged `.c4z` has not been installed on a controller.


derek-miller:
@svc-finitelabs please approve the PR if nothing is blocking

PR #95, opened 2026-08-09
chore: update to driver template v0.9.14

derek-miller:
Updates this repo to driver template `v0.9.14`.

## What comes in

The five vendored Lua libraries, refreshed from their release assets. Only the ones this repo declares in `vendor_modules` are touched.

| library | from | to |
|---|---|---|
| bitn | v0.6.0 | v0.6.1 |
| bthome | v0.1.2 | v0.1.3 |
| crypto | v0.1.0 | v0.2.0 |
| noiseprotocol | v0.6.0 | v0.6.2 |
| protobuf | v0.6.4 | v0.6.5 |

Two are correctness fixes rather than version bumps:

- **bitn v0.6.1** picked its native-operator implementation by testing whether `a & b` *parses*. LuaJIT rolling releases accept that syntax while returning signed 32-bit values, so the branch was taken on hosts needing the `to_unsigned()` normalisation it skips, and `arshift` returned wrong answers there. Control4 controllers run LuaJIT.
- **crypto v0.2.0** makes randomness strong-or-raise with no weak fallback. v0.1.0 seeded from `math.randomseed(os.time() + os.clock())`, worth roughly 20 bits at driver startup.

Plus the template delta since this repo last updated: `ParseDeviceIdList` numbering entries by position rather than by resolution, HTTP log redaction, and websocket fixes including a `%016X` length field that was previously space-padded into a malformed 14-byte field.

## Audit

I checked the changed API surface against every consumer in this repo rather than assuming the update was inert.

**`ParseDeviceIdList` index semantics changed** — it now counts every entry rather than only resolvable ones, so indices can have gaps. Across all ten driver repos, only `control4-home-connect` reads the third callback argument, at two sites in `power_strips.lua`. Both were checked: they use the same numbering on the publish and lookup sides, so they stay consistent. The only behavioural difference is that a transiently unresolvable outlet no longer renumbers the outlets behind it, which is the bug the change fixes.

**`http.lua`** additions are all `local`, and `redact()` is reachable only from `log:` calls, so it cannot alter request data.

**`websocket.lua`** — `Send(s, opcode)` defaults to the old hardcoded `0x81`, so callers are unaffected. The new `_netBindingByEndpoint` reuse changes binding allocation, but no driver references `netBinding` or `GetBindingAddress` outside `vendor/`. Every repo's vendored copy was pristine against its own template baseline, so nothing local was clobbered.

## Verification

`make test` green. Full suite run after the update, not just the template's own tests.

## This repo needed two fixes beyond the update

**Conflict markers in three vendored libraries.** `vendor/bitn.lua`, `vendor/crypto.lua` and `vendor/bthome.lua` all came out of the three-way merge with `<<<<<<<` markers. The copies here were the *same versions* the template shipped but *different amalg builds*, so line ordering differed enough that the merge could not reconcile them. Resolved by taking the v0.9.14 template copies wholesale, which are the published release artifacts.

This one mattered: `vendor/bitn.lua` did not parse at all with the markers in it, and this driver reaches bitn through both crypto and bthome.

```
luajit: error loading module 'bitn' ... vendor/bitn.lua:274: unexpected symbol near '<'
```

**Two manual integration tests started running.** `test_esphome_connection.lua` and `test_fatal_error.lua` both drive a real device and key off `ESPHOME_TEST_IP`; the latter called `os.exit(1)` when it was unset. This repo was on v0.9.12, which had no `test` target, so neither had ever run. The new target globs `test/test_*.lua` and picked them up. They now skip cleanly instead of failing the suite.

`104 passed, 2 skipped, 0 failures`.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Changes here will be overwritten by Copier; NEVER EDIT MANUALLY
-_commit: v0.9.12
+_commit: v0.9.14
 _src_path: git@github.com:finitelabs/control4-driver-template.git
 distributions: drivercentral oss
 github_org: finitelabs
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,7 @@ name: Build
 on:
   push:
     branches: [main, master]
+    tags: ['v*']
   pull_request:
     branches: [main, master]
 
@@ -51,6 +52,9 @@ jobs:
 
       - name: Initialize project
         run: make init
+
+      - name: Test
+        run: make test
 
       - name: Build
         run: make build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,11 +11,13 @@ permissions:
 jobs:
   release:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
 
     steps:
       - name: Wait for build and download oss artifacts
         run: |
           echo "Waiting for Build workflow to complete for $GITHUB_SHA..."
+          deadline=$(( $(date +%s) + WAIT_MINUTES * 60 ))
           while true; do
             run_id=$(gh run list \
               --repo "$GITHUB_REPOSITORY" \
@@ -27,6 +29,10 @@ jobs:
             if [ -n "$run_id" ]; then
               echo "Build run $run_id completed"
               break
+            fi
+            if [ "$(date +%s)" -ge "$deadline" ]; then
+              echo "::error::No completed Build run for $GITHUB_SHA after $WAIT_MINUTES minutes. Check that Build was triggered for this tag and is not queued behind another run."
+              exit 1
             fi
             echo "Build not finished yet, waiting 30s..."
             sleep 30
@@ -46,6 +52,7 @@ jobs:
             --dir dist/oss
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          WAIT_MINUTES: 20
 
       - name: Create release
         uses: softprops/action-gh-release@v2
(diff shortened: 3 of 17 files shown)


derek-miller:
Already fixed, and you are right about the audit sentence.

The dirty-tree failure was fixed before this review landed. `make build` was run and the reformatted vendored files committed, so CI is green on the current head. Your diagnosis matches exactly what I found independently: stylua over `vendor/`, the release assets being raw squish output, and the reformat being cosmetic.

## Your correction to the audit is right, and it is about the method

> Every repo's vendored copy was pristine against its own template baseline, so nothing local was clobbered.

That sentence is too broad and the method behind it does not generalise. Comparing a raw template blob to a repo blob is invalid for anything stylua reformats, which is every amalg-produced bundle. Any such comparison reports drift that is not drift.

I checked whether my specific comparison was valid or merely lucky, since those need different responses:

```
stylua(vendor/drivers-common-public/module/websocket.lua) == the committed file
```

Unchanged, so the check I actually ran holds for that file, because `websocket.lua` is hand-written and already in stylua form. But that is a property of that one file, not of the method, and the sentence claimed the general thing. The comparison has to be made after stylua, as you say.

This is also the real explanation for the merge conflicts in `control4-esphome#95`. I recorded those as "same versions but different amalg builds". They were the same content, one side stylua-formatted, which is exactly why a three-way merge could not reconcile them and why the diff looked enormous.

## The v0.7.0 to v0.8.0 history is the useful part

That the template shipped the formatted bundle through `v0.7.0` and the raw squish output from `v0.8.0` explains why this never surfaced before: earlier updates ran `make build` before committing and absorbed the reformat silently. I verified with `make test`, which does not format, so I never triggered it.

That looks like a template wart rather than a per-repo one. Every consuming repo now has to reformat on build to stay green, and the template's own CI cannot catch it because the template has no build. Worth deciding whether the template should commit these pre-formatted. I will raise it separately rather than widen this PR.

PR #99, opened 2026-08-27
Add support for climate, HVAC presets, scheduling and vane swing

McGlothlin-Inc:
_Summary_
Adds presets, preset scheduling, holds and vane (swing) control to the
esphome_climate driver, tested on a Mitsubishi mini-split over CN105 on OS
4.2.1, not only under the shim.

_Changes_
- PRESET_FIELDS_CHANGED, DYNAMIC_CAPABILITIES_CHANGED and
  ALLOWED_HOLD_MODES_CHANGED pushed on entity attach.
- driver.xml capability tags corrected (can_do_auto, per-mode setpoint
  resolution).
- Offline reporting via has_connection_status and CONNECTION.
- Water heater setpoint NaN handling; humidity now on binding 5011 (type 1),
  temperature on 5010.
- Rebased onto current main, which now provides the NaN-safe protobuf decoding,
  the NaN/infinity sentinels in the bridge serialization and the sensor
  non-finite handling this branch previously carried; those copies are dropped
  and the tests that cover them remain.

test/c4_shim.lua is template-owned and is the template's copy byte for byte, as of
v0.9.25. C4:ParseXml now ships in the template itself (control4-driver-template#62),
so the local patch this branch used to carry is gone.

_Breaking change_
Setpoint count now follows supports_two_point_target_temperature alone. Most
components declare a single target, so existing installs move to one setpoint,
and a preset saved with a heat/cool pair needs its setpoint entered again. This
warrants a major version.

_Design notes_
- thermostatV2 ignores the static preset_fields, can_preset and hold_modes tags,
  so the driver pushes them at runtime. can_preset_schedule is a change from the
  earlier revision, which left it static because the SDK documents only
  can_preset as dynamic: with the static tag alone the Schedule screen was
  absent on a controller running OS 3.3.3, and pushing CAN_PRESET_SCHEDULE
  beside CAN_PRESET restored it, so the driver now declares it statically and
  pushes it too; with the push in place the Schedule screen and presets are
  offered on both OS 3.3.3 and OS 4.2.1.
- Changed from this PR's earlier revision, which read SET_EVENTS and ran its own
  schedule clock: the proxy keeps the clock (full SET_EVENTS on each edit and
  connect; SET_EVENT for the preset in force on save, at each boundary where the
  preset changes, and on connect), and the driver applies SET_EVENT. A 60 second
  poll covers the one case the proxy does not announce, a boundary that re-selects
  the preset already in force, and acts on nothing else.
- The driver applies that preset, drops any standing hold, and persists the name
  so the connect repeat is ignored.

Device-side ESPHome presets are not mapped to proxy presets. The API publishes
which presets exist and which is active but never their values, and a device
keeps reporting its preset after the user overrides it, so a proxy preset
standing in for it would match permanently and a hold could never engage. This
was a deliberate decision, measured on hardware: the reason is a property of the
API, not an implementation gap.

_Tests_
- 249 assertions in test/test_climate_driver.lua, 5 in
  test/test_bridge_disconnect.lua, 14 in test/test_protobuf_float.lua, on the
  shared testlib harness, no device needed.
- Flagged, not fixed: test/test_esphome_connection.lua and
  test/test_fatal_error.lua call runEventLoop(), defined nowhere; the shim
  provides ShimRunEventLoop.

### DIFF
--- a/drivers/esphome_climate/driver.xml
+++ b/drivers/esphome_climate/driver.xml
@@ -112,8 +112,8 @@
         <field id="heat_setpoint_c" type="number" label="Heat Setpoint" min="4" max="32" res=".5"/>
         <field id="cool_setpoint_f" type="number" label="Cool Setpoint" min="50" max="99" res="1"/>
         <field id="cool_setpoint_c" type="number" label="Cool Setpoint" min="10" max="37" res=".5"/>
-        <field id="single_setpoint_f" type="number" label="Setpoint" min="40" max="90" res="1"/>
-        <field id="single_setpoint_c" type="number" label="Setpoint" min="4" max="40" res="1"/>
+        <field id="single_setpoint_f" type="number" label="Setpoint" min="39" max="90" res="1"/>
+        <field id="single_setpoint_c" type="number" label="Setpoint" min="4" max="32" res=".5"/>
         <field id="hvac_mode" type="list" label="HVAC Mode">
             <list>
                 <item text="Off" value="Off"/>
--- a/drivers/esphome_climate/driver.xml
+++ b/drivers/esphome_climate/driver.xml
@@ -219,7 +219,7 @@
       <id>5010</id>
       <facing>6</facing>
       <connectionname>Temperature</connectionname>
-      <type>6</type>
+      <type>1</type>
       <consumer>False</consumer>
       <audiosource>False</audiosource>
       <videosource>False</videosource>
@@ -235,7 +235,7 @@
       <id>5012</id>
       <facing>6</facing>
       <connectionname>Humidity</connectionname>
-      <type>6</type>
+      <type>1</type>
       <consumer>False</consumer>
       <audiosource>False</audiosource>
       <videosource>False</videosource>
--- a/drivers/esphome_climate/driver.xml
+++ b/drivers/esphome_climate/driver.xml
@@ -105,8 +105,48 @@
     <can_auto>True</can_auto>
     <has_humidity>False</has_humidity>
     <has_single_setpoint>False</has_single_setpoint>
-    <can_preset>False</can_preset>
-    <can_preset_schedule>False</can_preset_schedule>
+    <can_preset>True</can_preset>
+    <can_preset_schedule>True</can_preset_schedule>
+    <preset_fields>
+        <field id="heat_setpoint_f" type="number" label="Heat Setpoint" min="40" max="90" res="1"/>
+        <field id="heat_setpoint_c" type="number" label="Heat Setpoint" min="4" max="32" res=".5"/>
+        <field id="cool_setpoint_f" type="number" label="Cool Setpoint" min="50" max="99" res="1"/>
+        <field id="cool_setpoint_c" type="number" label="Cool Setpoint" min="10" max="37" res=".5"/>
+        <field id="single_setpoint_f" type="number" label="Setpoint" min="40" max="90" res="1"/>
+        <field id="single_setpoint_c" type="number" label="Setpoint" min="4" max="40" res="1"/>
+        <field id="hvac_mode" type="list" label="HVAC Mode">
+            <list>
+                <item text="Off" value="Off"/>
+                <item text="Heat" value="Heat"/>
+                <item text="Cool" value="Cool"/>
+                <item text="Auto" value="Auto"/>
+                <item text="Fan Only" value="Fan Only"/>
+                <item text="Dry" value="Dry"/>
+            </list>
+        </field>
+        <field id="fan_mode" type="list" label="Fan Mode">
+            <list>
+                <item text="Auto" value="Auto"/>
+                <item text="On" value="On"/>
+                <item text="Off" value="Off"/>
+                <item text="Low" value="Low"/>
+                <item text="Medium" value="Medium"/>
+                <item text="Middle" value="Middle"/>
+                <item text="High" value="High"/>
+                <item text="Focus" value="Focus"/>
+                <item text="Diffuse" value="Diffuse"/>
+                <item text="Quiet" value="Quiet"/>
+            </list>
+        </field>
+        <field id="swing" type="list" label="Swing">
+            <list>
+                <item text="Off" value="Off"/>
+                <item text="Vertical" value="Vertical"/>
+                <item text="Horizontal" value="Horizontal"/>
+                <item text="Both" value="Both"/>
+            </list>
+        </field>
+    </preset_fields>
     <has_temperature>True</has_temperature>
     <current_temperature_min_c>-50</current_temperature_min_c>
     <current_temperature_max_c>150</current_temperature_max_c>
@@ -137,7 +177,7 @@
     <has_remote_sensor>True</has_remote_sensor>
     <fan_modes/>
     <hvac_modes>Off,Heat,Cool,Auto</hvac_modes>
-    <hold_modes/>
+    <hold_modes>Off,Until Next</hold_modes>
   </capabilities>
   <proxies>
     <proxy proxybindingid="5001" primary="True" name="ESPHome Climate">thermostatV2</proxy>
--- a/drivers/esphome_climate/driver.xml
+++ b/drivers/esphome_climate/driver.xml
@@ -100,19 +100,19 @@
     <actions/>
   </config>
   <capabilities>
-    <can_heat>True</can_heat>
-    <can_cool>True</can_cool>
-    <can_auto>True</can_auto>
+    <can_heat>False</can_heat>
+    <can_cool>False</can_cool>
+    <can_auto>False</can_auto>
     <has_humidity>False</has_humidity>
     <has_single_setpoint>False</has_single_setpoint>
-    <can_preset>True</can_preset>
+    <can_preset>False</can_preset>
     <can_preset_schedule>True</can_preset_schedule>
     <preset_fields>
         <field id="heat_setpoint_f" type="number" label="Heat Setpoint" min="40" max="90" res="1"/>
         <field id="heat_setpoint_c" type="number" label="Heat Setpoint" min="4" max="32" res=".5"/>
         <field id="cool_setpoint_f" type="number" label="Cool Setpoint" min="50" max="99" res="1"/>
         <field id="cool_setpoint_c" type="number" label="Cool Setpoint" min="10" max="37" res=".5"/>
-        <field id="single_setpoint_f" type="number" label="Setpoint" min="39" max="90" res="1"/>
+        <field id="single_setpoint_f" type="number" label="Setpoint" min="40" max="90" res="1"/>
         <field id="single_setpoint_c" type="number" label="Setpoint" min="4" max="32" res=".5"/>
         <field id="hvac_mode" type="list" label="HVAC Mode">
             <list>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,35 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Removed
 -->
 
-<!-- #ifndef DRIVERCENTRAL -->
-
 ## Unreleased
 
+### Added
+
+- Added presets to the climate driver. A preset stores a setpoint, HVAC mode,
+  fan mode and vane position, and can be applied from the app or from
+  programming.
+- Added preset scheduling. Presets can be scheduled by weekday and time, and the
+  schedule survives a controller restart.
+- Added holds. Changing the thermostat by hand holds the new setting until the
+  next scheduled event, which then releases it.
+- Added vane control for climate devices that report swing modes, in the Extras
+  tab.
+
+### Changed
+
+- Climate devices that report a single setpoint now show one setpoint instead of
+  a heat and cool pair. Most heat pumps and mini splits work this way: they hold
+  one target and decide internally whether to heat or cool toward it, so the
+  pair could never be honored. Auto is unaffected. Any preset saved before this
+  release that carried separate heat and cool setpoints needs its setpoint
+  entering again.
+
 ### Fixed
+
+- Fixed the climate driver's temperature and humidity outputs appearing as audio
+  connections in Composer instead of control connections.
+
+<!-- #ifndef DRIVERCENTRAL -->
 
 - Fixed an automatic update sometimes leaving companion drivers on the previous
   version until the next update, which could make them stop responding in the
--- a/README.md
+++ b/README.md
@@ -906,7 +906,31 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ## Unreleased
 
+### Added
+
+- Added presets to the climate driver. A preset stores a setpoint, HVAC mode,
+  fan mode and vane position, and can be applied from the app or from
+  programming.
+- Added preset scheduling. Presets can be scheduled by weekday and time, and the
+  schedule survives a controller restart.
+- Added holds. Changing the thermostat by hand holds the new setting until the
+  next scheduled event, which then releases it.
+- Added vane control for climate devices that report swing modes, in the Extras
+  tab.
+
+### Changed
+
+- Climate devices that report a single setpoint now show one setpoint instead of
+  a heat and cool pair. Most heat pumps and mini splits work this way: they hold
+  one target and decide internally whether to heat or cool toward it, so the
+  pair could never be honored. Auto is unaffected. Any preset saved before this
+  release that carried separate heat and cool setpoints needs its setpoint
+  entering again.
+
 ### Fixed
+
+- Fixed the climate driver's temperature and humidity outputs appearing as audio
+  connections in Composer instead of control connections.
 
 - Fixed an automatic update sometimes leaving companion drivers on the previous
   version until the next update, which could make them stop responding in the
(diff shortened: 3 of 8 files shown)


derek-miller:
@svc-finitelabs can you review

drivers/esphome_climate/driver.xml
```diff
@@ -105,8 +105,48 @@
     <can_auto>True</can_auto>
     <has_humidity>False</has_humidity>
     <has_single_setpoint>False</has_single_setpoint>
-    <can_preset>False</can_preset>
-    <can_preset_schedule>False</can_preset_schedule>
+    <can_preset>True</can_preset>
```
derek-miller: I believe these are dynamic and should be set to false until the device connects and supports it. I would want the presets to work with the native esphome presets if possible. Otherwise if these are purely driver presets that we make up, then we need to make sure what we expose is supported by the device (eg. if we are allowing heat setpoint, does it even have heat?).

McGlothlin-Inc: Agreed, On it

derek-miller:
@svc-finitelabs can you build a few esphome dummy yamls with various feature and run them locally, install the updated drivers to the dev controller and test the new functionality and look for regressions. Keep them running and let me know and I can test from the navigator side of things if you give me a test procedure.

derek-miller:
@svc-finitelabs I ran all the tests and everything worked as described, though I think we need to determine a path forward to allowing the humidity connection to be exposed. What are the options for doing so?

McGlothlin-Inc:
Both patches are in, with tests. I also put the change on hardware, and it turned
up something that changes finding 1.

_What is committed_
Humidity renumbered to 5011 in `driver.xml` and `driver.lua`, and the swing
preset field gated on more than one mode. Two tests added, each failing without
its fix: reverting the swing gate fails the withheld-swing assertion, and putting
humidity back on 5012 fails the binding assertions. Fan mode left alone, per your
reasoning.

_What the hardware shows_
Two controllers, OS 4.2.1 and OS 3.3.3, same head, same parent driver version.
Identical results on both.

On mainline, freshly installed, the child item reports only 5001 and 5002. The
proxy item reports the union, and that is where the statics are:

```
Audio    5010 Temperature, 5012 Humidity   owner=proxy
Control  1    Temperature                  owner=proxy
```

5012 is not deleted. It exists, typed Audio because mainline declares it
`<type>6</type>`, and it is owned by the proxy item rather than the driver.

That is what reconciles our two readings. Both are correct and neither view is
wrong — they are different items. Reading the child, these rows are absent on
every install and every OS, which looks exactly like a deletion. Reading the
proxy union, they are present.

The ownership rule turns out to explain the rest of it:

| declared type | shows as | owned by |
|---|---|---|
| `<type>6</type>` | A/V | proxy item |
| `<type>1</type>` | Control | driver item |

_Two consequences worth knowing before merge_

**Updating in place leaves the old rows behind.** Changing 6 to 1 does not
convert a record, it moves ownership to a different item and abandons the old
one. An existing install shows a phantom Temperature and Humidity under A/V
alongside the correct Control rows. A fresh install does not.

They clear on the next controller restart — measured separately on both OS
versions, not generalised from one. Nothing is written to the project, and no
manual cleanup is needed. Worth a changelog line, but it is not a blocker.

A driver-side migration cannot help here. I built a variant calling
`C4:RemoveDynamicBinding(5012)` at init and it had no effect, which follows from
the table above: that call runs in the driver's context and the stale record
belongs to the proxy.

**The duplicate Temperature is inherent, not an update artifact.** On a clean
install of this branch, Control shows two Temperature rows: the proxy's own
`binding 1` and the driver's static 5010, same reading. On mainline this was
invisible because 5010 sat under A/V; moving it to Control puts it next to the
proxy's.

Which is your own argument for the humidity static, applied to temperature — the
proxy already provides it. The counter-argument was that installs may have 5010
wired, and that is worth weighing against the fact that on mainline 5010 was an
A/V connection, so anything bound to it was bound as A/V. Happy to drop the
driver's 5010 in this PR if you want the single row, or leave it and revisit.

_Not tested_
Whether humidity values actually publish. The head here does not report humidity,
so the proxy's capability-gated humidity output never appears and the path cannot
be exercised on the hardware I have. Your mocks are the better instrument for
that one.

derek-miller (commented):
Reviewed at `cd500f1`, this time by exercising the driver rather than the diff: I ran the branch's own `driver.lua` under `test/c4_shim.lua` with targeted probes. Five of the findings below were reproduced that way, not inferred. The prior rounds were thorough about what they exercised, but the hold state machine, the reload/restore ordering, empty presets, and water heaters were never on the rig.

## Blockers

### 1. An empty preset matches every state and disables holds

`parsePresetFields` drops any field with an empty value (`driver.lua:1148`) and returns `{}` for an unparsable `preset_fields` attribute, and `RFP.SET_PRESETS` stores the result without a guard. Every check in `matchPreset` (1303-1368) skips a nil field, so an empty preset falls through to `return true` at 1367 against any state.

Reproduced under the shim: with an all-empty preset named `Empty` alongside a real one, an arbitrary state report announced `PRESET_CHANGED NAME=Empty` — which preset wins is `pairs` order over the names, the same lottery the seeding design was rejected for. And with `SCHEDULED_PRESET = "Empty"`, no divergence ever raised a hold: `matchPreset` cannot see divergence from a preset that constrains nothing, so the hold feature is dead for that schedule.

`applyPreset` already has a "defines no usable fields" branch (1286-1288), so field-less presets are an anticipated input. `SET_PRESETS`/`matchAnyPreset` need the same treatment: skip (or refuse to store) a preset whose parsed field table is empty.

### 2. A user-raised hold is cancelled by the next state report

`RFP.SET_MODE_HOLD`'s non-Off branch (1593-1601) records nothing about user intent; `reconcileHold` (1403-1412) runs on every climate state report and unconditionally forces `setHoldMode("Off")` whenever the state matches `SCHEDULED_PRESET`.

Reproduced: raise "Until Next" from the UI without changing anything, deliver one ambient-temperature push (they arrive every few seconds), and `HOLD_MODE_CHANGED {MODE="Off"}` is emitted immediately. The user's explicit hold survives only until the device next reports. Related: `fireScheduledEvent` (1491-1502) never consults `HOLD_MODE`, so even a surviving user hold would not block the schedule — the hold only works in the divergence direction. `reconcileHold` needs to distinguish "hold I raised from divergence" from "hold the user asked for".

### 3. Water heaters get a preset UI that can't work

`sendCapabilities` sets `CAN_PRESET = true` unconditionally (547), including for water heaters, but `PRESET_FIELDS_CHANGED` is gated on `not entity.is_water_heater` (554) — so a water heater's preset editor is served the static climate-shaped template from driver.xml (heat/cool setpoints, HVAC modes, swing). Applying one is a wire-level no-op: `applyPreset` ends in a command-less `sendClimateCommand(body)`, the parent defaults that to `water_heater_command` (`src/esphome/entities/water_heater.lua:71-72`), and the climate-shaped body serializes against `WaterHeaterCommandRequest` with `has_fields = 0`, so the device applies nothing. Matching is gated off for water heaters in UPDATE_STATE (1860-1863), so it's never announced either. Gate `CAN_PRESET` and the `SET_PRESET`/`SET_PRESETS` handlers on `not entity.is_water_heater`.

## Should fix — scheduling correctness (fine as an immediate follow-up if you'd rather)

**4. A schedule boundary crossed before presets arrive is lost for a week.** `SCHEDULE` is persisted; `PRESETS` is not, and the proxy only resends `SET_PRESETS` after the `CONNECTION` notify at the end of `sendCapabilities` (689) — i.e. after the ESPHome device connects. Reboot shortly before an event (or with the device offline): the restored timer fires, `fireScheduledEvent` hits `PRESETS[event.preset] == nil`, warns, and re-arms — and `secondsUntilEvent`'s `delta < 1` clause (1116) resolves the just-missed event ~7 days out. Nothing retries when presets land: the `SET_PRESETS` re-apply path needs `signatureBefore ~= nil`, which can't hold after a reload. The device sits on stale settings until the next boundary.

**5. `SET_EVENT` before `SET_PRESETS` leaves holds dead.** The guard at 1564-1567 drops the record when `PRESETS[name]` is nil, nothing sets `SCHEDULED_PRESET` retroactively when presets arrive, and `reconcileHold` no-ops on nil. If the proxy's post-CONNECTION resend order puts `SET_EVENT` first — the driver has no contract saying it can't — divergence holds and hold release are inert until the next boundary fires. Both 4 and 5 share the root cause: `PRESETS` is the one piece of schedule-critical state neither persisted nor re-derived, and both consumers drop instead of deferring.

**6. Of two events at the same weekday/time, the second never fires.** Verified by probe: the strict `<` tie-break (1127) picks the first in `SCHEDULE` order each arm; after it fires, both events compute `delta < 1` → +7 days, and the first wins the tie again, every week. `fireScheduledEvent` applies only the single selected event.

**7. Releasing a hold with no schedule leaves `HOLD_PRESET` set.** The Off branch clears it only inside `if SCHEDULED_PRESET ~= nil` (1589-1590), while `RFP.SET_PRESET` sets `HOLD_PRESET = name; SCHEDULED_PRESET = nil` (1483-1484). Reproduced: apply a preset, release the hold, edit that preset in the app → the device is commanded with the edit via `activeAfter = HOLD_PRESET or ...` (1459) despite the release. The empty-NAME path at 1477-1479 clears unconditionally; the asymmetry looks like an oversight.

**8. The hold flaps at every scheduled boundary if a report lands in the command window.** Same race the PR engineered away for `PRESET_CHANGED`: a stale report between `fireScheduledEvent`'s command and the device's confirmation makes `reconcileHold` emit `Until Next`, then `Off` on confirmation — two spurious programmable `HOLD_MODE_CHANGED` events per boundary. Reproduced under the shim; `PRESET_CHANGED` correctly did not flap in the same run.

## Two plausible, one worth a hardware minute

**9. Match tolerance vs device quantization.** `matchPreset` uses ±0.25 °C (1338) but the template authors at 0.5 °C resolution and `applyPresetSetpoints` never snaps to `visual_target_temperature_step`. Probed on a step-1 entity: a 21.5 °C scheduled preset echoed back as 22 raises "Until Next" permanently and un-highlights the preset. The hardware pass didn't hit it because the captured preset was an integer. Snapping preset setpoints to the entity step before sending (or widening the tolerance to the authoring resolution) closes it.

**10. All capabilities now start False.** `can_heat`/`can_cool`/`can_auto`/`can_preset` are static False and only `DYNAMIC_CAPABILITIES_CHANGED` raises them, which requires the device online. If the proxy doesn't persist dynamic caps, a Director restart with the device offline shows a thermostat that can't heat, cool, or auto until reconnect. Every in-repo link is verified; the proxy's persistence isn't observable from here. Worth one test: restart Director with a mock powered off and open the thermostat in Navigator.

## Non-blocking cleanup

- `xmlAttr` (388-397) re-implements `XMLEncode` from `drivers-common-public/global/lib.lua:462`, which is already required at line 8.
- Every device reconnect re-runs `sendCapabilities` → CONNECTION notify → `SET_EVENTS` → an unconditional `persist:set("Schedule", ...)` (1552): an identical-content flash write per reconnect on a flaky device. `Persist:set` doesn't dedupe.
- `matchAnyPreset` + `reconcileHold` run N+1 full preset comparisons on every state push, including temp-only pushes that can't change any outcome (`matchPreset` never reads `current_temperature`). A signature over the compared fields would skip nearly all of it.
- The swing `EXTRAS_STATE_CHANGED` echo (1865-1885) fires for `#supported_swing_modes > 0` while the selector setup requires `> 1`, so a lone-Off device gets state changes for an object it never declared, on every push, with no transition guard. One correction to the earlier thread: this isn't pre-existing — base has no swing code at all, so it's in scope for this PR.

Items 1-3 I'd like fixed on this branch; 4-8 can ride a fast follow-up if you prefer, your call. The feature work itself remains solid — these are all in the corners the rig didn't reach.

derek-miller:
@svc-finitelabs please review

McGlothlin-Inc:
Item 10, from a run of unplug, reboot while unplugged, re-plug, watching a touchscreen:

Unplugged, no restart: temperature "--", all controls withheld, IS_CONNECTED 0.

After the restart with the device still offline: OnDriverLateInit sent CONNECTION CONNECTED=false and IS_CONNECTED read 0, yet the screen showed "--" with the full mode list (Off/Heat/Cool/Auto/Fan Only/Dry), the hold list (Off/Until Next), and a Schedule screen with no preset control; HVAC_MODES_LIST and HOLD_MODES_LIST still held the pre-restart lists.

On reconnect: SET_EVENTS and SET_EVENT arrived, the scheduled preset applied, hold modes republished, temperature and controls returned, presets and schedule intact; the Swing entry under Extras appeared only after re-entering the screen.

So, the proxy persists the published hvac_modes and hold_modes across a restart, so can_heat/can_cool/can_do_auto never narrow the mode list and False versus absent has no visible effect; withholding follows the live CONNECTION notification only, and presets wait for reconnect because can_preset is raised on attach.

McGlothlin-Inc:
Both call sites fixed as suggested: `SET_PRESETS` and `SET_EVENTS` now pass `Select(tParams, "XML") or ""`, so a frame without an XML param reaches the existing "warn and keep what we have" branch instead of raising inside the proxy handler. `C4:ParseXml("")` returns `nil` on a controller, measured here as well as on your side, so the guards below each call work as written.

Test added: both handlers called with no XML param return normally and republish nothing. Against the unfixed driver under a shim with Director's argument checks it fails at both sites with `strXml should be a string`, so it is load-bearing.

`test/c4_shim.lua` is now the template's merged copy, byte for byte, with the local-patch banner gone. That puts the strict argument checks and the measured parsed result into this repo's tests today, and the next `copier update` finds the file already identical, so the clobber you flagged has nothing left to clobber.

Suite at the new head: `test_climate_driver.lua` 252 passed, 0 failed; all 19 files green; `stylua --check` with the CI flags clean. Not rebased: the branch is mergeable against `main` as it stands and the shim file now agrees with the template, so a rebase would only re-arm the gate for no change in content. Two commits, plain fast-forward on the fork.

McGlothlin-Inc:
Merged `main` into the branch after #123 landed, taking the incoming side of every conflict, then one fix on top for what that merge broke here.

The merge brought a second helper for fields protobuf leaves off the wire, and the three readings it took from `main` lost the non-finite filter this branch already had: a NaN or infinite current temperature, humidity or humidity target reached the proxy as a number again (8 assertions in `test_climate_driver.lua` go red on the bare merge). The fix keeps one helper under `main`'s name with both rules, absent means zero for a dimension the entity declares, and a non-finite value means nothing yet, and reads every field through it.

`test_climate_zero_fields.lua` cuts `UPDATE_STATE` out of the source and runs it in a stub environment. This branch's handler also reaches the swing table and the hold hooks, so the cut now takes the table and the environment stubs the hooks; the vacuity guard names the last cut piece instead of a fixed index. README regenerated from the merged changelog, byte-identical on a second run.

At the new head: all 23 suites green, `test_climate_zero_fields.lua` 39 passed, `test_climate_driver.lua` 252 passed, `stylua --check` with the CI flags clean.

McGlothlin-Inc:
Confirming `94a24e1` from this side. I had resolved the same merge locally before your push landed, and the two trees are byte for byte identical, so nothing further is coming from this branch.

Hardware run on that exact tree, on a controller with a Mitsubishi mini-split behind the ESPHome bridge: both drivers built with the repo's own pipeline and installed, then Director restarted so the proxy re-reads `driver.xml`.

- Both drivers reconnected at the new versions. The proxy's loaded capabilities carry `hvac_states` from #127 alongside this branch's `hvac_modes` with Fan Only and Dry and its two hold modes.
- Current temperature reads as a number on both scales, no NaN, on the merged `stateNumber` path.
- The proxy's repeated schedule event after reconnect was logged as already in force and left alone.
- `HVAC_STATE` stays empty on this unit, and that is correct rather than a gap: the ESP's own event stream for this climate entity carries no `action` field, so the driver has nothing to send. #127's list is in place for a unit that does report it.
- A hold standing before the restart reads Off after it, which is the documented reload behaviour: a manual hold is memory-only and the first reconcile after a reload clears it when the state matches the scheduled preset.

Suite at the same tree here: 29 files, 1464 passed, 0 failed, `stylua --check` clean, README identical to a fresh render.

One observation, not a request: the proxy sends the full schedule list to the driver on every client connection to Director, not only when the driver itself connects. The driver deduplicates by signature, so it is harmless, but it explains why "Schedule updated" appears in the log whenever a client attaches.

derek-miller (approved):
Approved after a full review of the diff and its history, the gates run locally at each head, and a behavioural pass on the dev controller.

What I verified:
- The whole driver diff and all 40 review rounds; presets, scheduling, holds and swing on top of the proxy-driven design.
- `make check` / `make test` (29 suites, 1551 assertions, 0 failed) / full `make build`, clean tree, at this head.
- Every defect the dev-controller pass turned up is fixed on the branch, each with a revert-checked test: the Permanent hold downgrade, the double-command at a preset-changing boundary, the swallowed hold after it, presets forgotten with no schedule, the redundant CONNECTION announce, the per-report flash write, and hold state not surviving a reload. The last was fixed by persisting the hold; case 13 re-ran on the rig with zero device commands across the reload and the deferred preset applied once on release.

Two things measured but not on the branch, tracked for the follow-up (DRV-133), neither a merge blocker:
- A Director restart with a mock offline and one preset configured, to close the last unmeasured corner of the reload path from the UI side rather than the driver sandbox.
- The cleanup: move presets/scheduling/holds into their own module (removes the forward declarations), drop the learned hold wording, delete the inert static preset_fields block.

Issue #103, opened 2026-09-06
BUTTON object direction and feedback


krishy2:
I suppose the BUTTON objects should be also available in inverted direction. So if there is a button in the esphome that should act as BUTTON connection in C4 (somehow it is shown Input) to operate as a keypad and could be linked to other C4 lights.

Examle: 3 gang switch, 1 and 2 are used locally as toggle relays and can be toggled from the direction of C4, but the 3rd can toggle other C4 lights in the system.

Also the color feedback should be sent back to esphome somehow.

derek-miller:
Thanks for writing this up. There are two separate things tangled together here, and I want to
make sure we build the right one, so let me split them and then ask you for one thing.

**On the direction of the existing Button connections**

I checked this on a live controller against a mock ESPHome device with a `button` entity, and
the connection is created as a Control **Output**, not an Input. That matches how Control4
models button links, which is the opposite of what most people expect:

- A keypad's buttons are declared `consumer=True`, so they appear as Control **Inputs**
  (`control4_keypad_kpz-31-w.c4i`).
- A load's button links are declared `consumer=False`, so they appear as Control **Outputs**
  (`combo_dimmer.c4i`, "Top/Bottom/Toggle Button Link").

So "Output" here means "something a keypad button can be pointed at", which is the correct role
for an ESPHome `button` entity. An ESPHome `button` is an actuator you press *from* the
controller (the same thing HA's `button.press` service does), not a physical paddle that reports
presses. If you bind a Control4 keypad button to it, pressing the keypad presses the ESPHome
button. That part is working as intended and we do not want to flip it.

**On the thing you actually want**

Your 3-gang example is the real request, and it is a genuine gap: a *physical paddle* on the
ESPHome device cannot currently be wired straight to other Control4 loads the way a keypad
button can. Today a paddle shows up as a `binary_sensor` and gets a `CONTACT_SENSOR` connection
plus a driver variable, so you can already make gang 3 toggle another Control4 light through
Programming ("when the paddle closes, toggle Light X"). What is missing is the direct
keypad-style connection with no Programming, plus the LED feedback you mentioned.

That is buildable. We already do exactly this in the BTHome and SwitchBot sub-drivers, which
create `BUTTON_LINK` connections in the keypad direction and send `DO_PUSH`/`DO_RELEASE`/
`DO_CLICK` to whatever load they are bound to.

**What would help**

[...]

PR #104, opened 2026-09-06
fix: thermostat scale, thermostat online status, and light dimming caps

### LINKED ISSUE #100: Project is in C, but thermostat gives F
Control4 project is in Celsius, radio buttons are greyed out, Thermostat shows up in F.
C4 ver: 4.2.1

### LINKED ISSUE #101: Feature req - status display
In Control4 on the extras tab the driver status could be displayed to check if the device is online.

### LINKED ISSUE #102: Light_v2 driver error
Image

derek-miller:
Closes #100, closes #101, closes #102.

Each root cause was reproduced and verified against a live CORE 1 with host-mode
ESPHome mocks, not inferred from the reports.

## #100 - thermostat shows Fahrenheit in a Celsius project

Not `can_change_scale`, as suggested on the issue. The `thermostatV2` proxy keeps
its **own** per-thermostat display scale that defaults to Fahrenheit and never
consults `project_settings.scale`; the driver never told it otherwise.

Verified live: sending `SCALE_CHANGED { SCALE = "C" }` flips proxy variable 1100
`SCALE` to `CELSIUS` and recomputes `DISPLAY_TEMPERATURE` from 73 to 22.5.

The driver now reads the project scale with `C4:GetTemperatureScale()` and reports
it on connect, and implements `SET_SCALE` so an installer can override it for a
single thermostat. The override is persisted and wins over the project scale, so a
reconnect does not stomp it (verified). `can_change_scale` is now True: ESPHome is
always Celsius internally, so the scale is purely what Control4 displays and the
driver can always change it.

Note the `SCALE` param carried on `TEMPERATURE_CHANGED` and the setpoint notifies
is a different thing - it declares the unit of the value being sent - and is
unchanged.

## #101 - no way to see whether a device is online

Two parts.

**A real bug.** `thermostatV2` has **no `ONLINE_CHANGED`** - the string is absent
from `/control4/drivers/thermostatV2.c4l` - so the notification the climate driver
was sending was a silent no-op and a thermostat stayed shown as connected forever
after its device dropped. It reports reachability through `CONNECTION { CONNECTED }`,
which drives proxy variable 1112 `IS_CONNECTED`. Fixed, and paired with the
`has_connection_status` capability.

Verified end to end: killed the climate mock, `IS_CONNECTED` now goes to 0.

`light_v2` and `fan` do implement `ONLINE_CHANGED` and already worked (confirmed a
light goes offline correctly). The `lock` proxy implements neither; its only
offline-ish signal is `LOCK_STATUS = IS_FAULT`, so that one is left alone and the
docs say so.

**A documentation gap.** Every driver has published a `Connected` BOOL variable
since v20260802, which is exactly what the reporter wanted, but it appeared in no
documentation anywhere. Now documented in the main driver's Driver Variables table
and in a Programming Variables section on each sub-driver.

## #102 - light_v2 color rate exception and greyed dimming

**Color rate exception.** The color rate bounds and defaults now ride every
`DYNAMIC_CAPABILITIES_CHANGED` instead of only when the light has color, so there
is no window in which a consumer sees a color rate control with no range - the
shape of the `defaultColorRate ... value 0.750 not between 0.000 and 0.000`
exception. Confirmed by hand that the light properties dialog now opens in
Composer without throwing.

**Greyed dimming.** Composer's test panel gates its dimming controls on the static
driver.xml capability **AND** the dynamic one, so the conservative baseline in
0141b2c left an installer unable to test dimming on a light that really is
dimmable. Measured on a CORE 1 by deploying two builds differing only in the
static values and binding each to the same ESPHome entities:

```
static  dynamic   GET_SETUP   Composer properties   Composer test panel
False   true      True        enabled               greyed
False   false     False       greyed                greyed
True    true      True        enabled               enabled
True    false     False       greyed                greyed
```

Composer's properties tab and Navigator both follow the dynamic value alone; only
the test panel ANDs the two. So a static True costs nothing on a light that is not
dimmable - it reports a dynamic `dimmer=false` and is correctly greyed everywhere,
and slider-less in the app - and it restores the test panel on one that is. The
app slider was confirmed working throughout, so this was never customer-facing.

The dimming tier therefore moves to True: `dimmer`, `set_level`, `ramp_level`,
`supports_target`, `supports_brightness_stop`. The color tier stays False - a full
colour light with `supports_color` and `supports_color_correlated_temperature`
declared False came up with its colour picker and colour temperature control
usable, so the AND does not apply there.

Leaving the colour tier alone also limits the blast radius of DRV-113, the
consumer-side bug where a static value chained with `or` cannot be told apart from
an undeclared one. Home Connect reads `dimmer`, `set_level`, `ramp_level`,
`supports_color` and `supports_color_correlated_temperature` that way; it does not
read `supports_target` or `supports_brightness_stop`, so those two are free.

Two corrections to earlier reasoning in this PR's history, recorded because both
were asserted in code comments and are wrong:

- There is no `color_rate_default` / `brightness_rate_default` capability in the
  proxy. A fresh install declaring `<color_rate_default>0</color_rate_default>`
  still seeds `c4type::Light Color::rateDefault = 750` in `state.db`. The entries
  are kept because Composer reads the static capability list directly, but the
  proxy ignores them.
- A capability omitted from `DYNAMIC_CAPABILITIES_CHANGED` is not forced to 0;
  `color_rate_max` stayed at 65535 for a light that sent no color capabilities.

Also tested and **not** reproduced: the documented note that setting `dimmer=true`
makes the proxy re-read `ramp_level` from the c4i and clobber a `ramp_level=true`
sent in the same notify. One atomic notify reaches the correct state, and sending
`dimmer=true` alone raises only `dimmer`, so a split send gains nothing.

## #103 - not addressed here

The reporter's premise does not hold: an ESPHome `button` entity's `BUTTON_LINK`
is created as an **Output**, not an Input, which is the correct load-side role
(compare `combo_dimmer.c4i`, where a load's button links are `consumer=False`,
against `control4_keypad_kpz-31-w.c4i`, where a keypad's buttons are
`consumer=True`). The underlying request - a physical paddle on the device driving
Control4 loads directly - is a genuine gap, but it depends on which entity type
their paddle actually is. Awaiting their YAML.

`test/dummy_buttons.yaml` lands here as groundwork: a 3-gang switch mock covering
`switch`, `binary_sensor`, `event`, `button` and a status-LED `light`, which is
what established the binding directions above.

## Test plan

- [x] `make test` - 311 assertions, 0 failures
- [x] Merged into `refs/pull/99/head`: climate harness 135 passed, 0 failed (0/44 without the shim stub)
- [x] `make check` - formatting clean
- [x] `make build` - full build including docs and README regeneration
- [x] Thermostat scale follows the project setting on connect (live)
- [x] `SET_SCALE` override persists and survives a reconnect (live)
- [x] Thermostat `IS_CONNECTED` drops to 0 when the ESPHome device goes away (live)
- [x] Light properties dialog opens in Composer without the color rate exception
- [x] Dynamic capabilities override static in both directions (live, both XML variants)
- [x] Full static/dynamic matrix checked in Composer across four devices
- [x] Composer test panel dimming live on a dimmable light, greyed on an on/off one
- [x] Colour picker and CCT usable with the colour tier left False
- [x] Brightness slider works in the Control4 app throughout
- [x] Binary light still reports `dimmer=False` in `GET_SETUP` with static True
- [x] RGBCT light reports colour, CCT and a 2700-6500K range from device discovery

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
   ESPHome device went offline.
 - Fixed an "Error setting default color rate from driver" message in Composer
   when opening the properties of an ESPHome light.
+- Fixed Composer's test panel greying out the dimming controls for ESPHome
+  lights that do support brightness. The Control4 app was always able to dim
+  these lights; only Composer's own test controls were affected.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,8 +35,8 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed an "Error setting default color rate from driver" message in Composer
   when opening the properties of an ESPHome light.
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
-  lights that do support brightness. The Control4 app was always able to dim
-  these lights; only Composer's own test controls were affected.
+  lights that do support brightness. Dimming from the Control4 app was never
+  affected.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,20 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed an automatic update sometimes leaving companion drivers on the previous
   version until the next update, which could make them stop responding in the
   meantime.
+- Fixed thermostats and water heaters always showing Fahrenheit. They now follow
+  the project's temperature scale, and the Celsius/Fahrenheit setting in
+  Composer can be used to override it for an individual thermostat.
+- Fixed thermostats and water heaters staying shown as connected after the
+  ESPHome device went offline.
+- Fixed the dimming controls being greyed out in Composer for ESPHome lights
+  that do support brightness.
+- Fixed an "Error setting default color rate from driver" message in Composer
+  when opening the properties of an ESPHome light.
+
+### Changed
+
+- Documented the `Connected` variable that every driver publishes, so it can be
+  used in Programming to show whether a device is online.
 
 <!-- #endif -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,6 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
   Composer can be used to override it for an individual thermostat.
 - Fixed thermostats and water heaters staying shown as connected after the
   ESPHome device went offline.
-- Fixed the dimming controls being greyed out in Composer for ESPHome lights
-  that do support brightness.
 - Fixed an "Error setting default color rate from driver" message in Composer
   when opening the properties of an ESPHome light.
 
--- a/README.md
+++ b/README.md
@@ -923,8 +923,6 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
   Composer can be used to override it for an individual thermostat.
 - Fixed thermostats and water heaters staying shown as connected after the
   ESPHome device went offline.
-- Fixed the dimming controls being greyed out in Composer for ESPHome lights
-  that do support brightness.
 - Fixed an "Error setting default color rate from driver" message in Composer
   when opening the properties of an ESPHome light.
 
--- a/README.md
+++ b/README.md
@@ -435,15 +435,22 @@ supported ESPHome entity. Use this reference for Control4 programming.
 ### Driver Variables
 
 In addition to per-entity variables, the driver publishes the following
-device-level variables sourced from the ESPHome device info. They mirror the
-matching read-only Device Info properties.
+device-level variables. `Connected` tracks the live connection; the rest mirror
+the matching read-only Device Info properties.
 
-| Variable     | Type   | Description                                  |
-| ------------ | ------ | -------------------------------------------- |
-| Name         | STRING | Friendly name reported by the ESPHome device |
-| Model        | STRING | Device model string reported by ESPHome      |
-| Manufacturer | STRING | Manufacturer string reported by ESPHome      |
-| MAC Address  | STRING | ESPHome device MAC address                   |
+| Variable     | Type   | Description                                    |
+| ------------ | ------ | ---------------------------------------------- |
+| Connected    | BOOL   | True while the driver is talking to the device |
+| Name         | STRING | Friendly name reported by the ESPHome device   |
+| Model        | STRING | Device model string reported by ESPHome        |
+| Manufacturer | STRING | Manufacturer string reported by ESPHome        |
+| MAC Address  | STRING | ESPHome device MAC address                     |
+
+> **Tip:** Every sub-driver publishes its own `Connected` variable too, tracking
+> the entity it is bound to. Use it in Programming to drive a custom Navigator
+> element, send a notification, or trigger anything else when a device drops
+> off. Thermostats, lights and fans additionally report their connection state
+> to their proxy, so Navigator greys them out on its own.
 
 ### Variables by Entity Type
 
@@ -911,6 +918,20 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed an automatic update sometimes leaving companion drivers on the previous
   version until the next update, which could make them stop responding in the
   meantime.
+- Fixed thermostats and water heaters always showing Fahrenheit. They now follow
+  the project's temperature scale, and the Celsius/Fahrenheit setting in
+  Composer can be used to override it for an individual thermostat.
+- Fixed thermostats and water heaters staying shown as connected after the
+  ESPHome device went offline.
+- Fixed the dimming controls being greyed out in Composer for ESPHome lights
+  that do support brightness.
+- Fixed an "Error setting default color rate from driver" message in Composer
+  when opening the properties of an ESPHome light.
+
+### Changed
+
+- Documented the `Connected` variable that every driver publishes, so it can be
+  used in Programming to show whether a device is online.
 
 ## v20260816 - 2026-08-16
 
--- a/README.md
+++ b/README.md
@@ -925,6 +925,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
   ESPHome device went offline.
 - Fixed an "Error setting default color rate from driver" message in Composer
   when opening the properties of an ESPHome light.
+- Fixed Composer's test panel greying out the dimming controls for ESPHome
+  lights that do support brightness. The Control4 app was always able to dim
+  these lights; only Composer's own test controls were affected.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -926,8 +926,8 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed an "Error setting default color rate from driver" message in Composer
   when opening the properties of an ESPHome light.
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
-  lights that do support brightness. The Control4 app was always able to dim
-  these lights; only Composer's own test controls were affected.
+  lights that do support brightness. Dimming from the Control4 app was never
+  affected.
 
 ### Changed
 
--- a/drivers/esphome/www/documentation/index.md
+++ b/drivers/esphome/www/documentation/index.md
@@ -535,15 +535,22 @@ supported ESPHome entity. Use this reference for Control4 programming.
 ### Driver Variables
 
 In addition to per-entity variables, the driver publishes the following
-device-level variables sourced from the ESPHome device info. They mirror the
-matching read-only Device Info properties.
+device-level variables. `Connected` tracks the live connection; the rest mirror
+the matching read-only Device Info properties.
 
-| Variable     | Type   | Description                                  |
-| ------------ | ------ | -------------------------------------------- |
-| Name         | STRING | Friendly name reported by the ESPHome device |
-| Model        | STRING | Device model string reported by ESPHome      |
-| Manufacturer | STRING | Manufacturer string reported by ESPHome      |
-| MAC Address  | STRING | ESPHome device MAC address                   |
+| Variable     | Type   | Description                                    |
+| ------------ | ------ | ---------------------------------------------- |
+| Connected    | BOOL   | True while the driver is talking to the device |
+| Name         | STRING | Friendly name reported by the ESPHome device   |
+| Model        | STRING | Device model string reported by ESPHome        |
+| Manufacturer | STRING | Manufacturer string reported by ESPHome        |
+| MAC Address  | STRING | ESPHome device MAC address                     |
+
+> **Tip:** Every sub-driver publishes its own `Connected` variable too, tracking
+> the entity it is bound to. Use it in Programming to drive a custom Navigator
+> element, send a notification, or trigger anything else when a device drops
+> off. Thermostats, lights and fans additionally report their connection state
+> to their proxy, so Navigator greys them out on its own.
 
 ### Variables by Entity Type
 
--- a/drivers/esphome_bluetooth_coordinator/www/documentation/index.md
+++ b/drivers/esphome_bluetooth_coordinator/www/documentation/index.md
@@ -529,6 +529,12 @@ ID for rooms) to avoid conflicts.
 Variable names for per-device and per-room variables include a unique suffix
 (MAC address for devices, room ID for rooms) to avoid conflicts.
 
+### Driver Variables
+
+| Variable  | Type | Description                                                     |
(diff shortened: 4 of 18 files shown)


derek-miller:
@svc-finitelabs review now

derek-miller:
Thanks, this was a genuinely useful three rounds. Pushed `9c34c3e` addressing most of it. Taking the points in order of how much they mattered.

## Fixed

**The shim gap (round 1, still open through round 3).** Correct and mine to fix, since this branch introduced the call. I reproduced it both ways rather than taking the number: merged into `refs/pull/99/head` in a scratch worktree, `0 passed / 44 failed` with `attempt to call method 'GetTemperatureScale' (a nil value)`, and `135 passed / 0 failed` with the stub. Removing only the stub from that same merge puts it back to 0/44. Added returning `"FAHRENHEIT"` for the reason you gave: the full word exercises `normalizeScale` rather than accommodating it.

**The rate defaults are inert on the dynamic side too.** You are right, and this is the better catch of the two. I had only established that the *static* entries were ignored, and kept the dynamic ones on the assumption they were the working half. They are not in either list, so `clampRate()` and both persist lookups were feeding keys nothing reads. Removed from the payload, which also removes the duplicated "color falls back to brightness" rule you flagged separately, since `getDefaultColorRate()` already encodes it. The static XML entries stay, but the comment now says they are parity with Chowmain's ESPHome light rather than claiming to be the mechanism, and points at `color_rate_min`/`color_rate_max` as what actually does the work.

**`driver.lua:362` contradicting the XML.** Yes. I flipped the XML in `fdbd3af` and left the Lua comment describing the reverted contract. Rewritten to describe the case that now matters, a light without brightness being corrected back down.

**Project scale needing a reconnect.** Fixed rather than accepted. `sendDisplayScale()` is now re-evaluated on every state update and only notifies when the value changed, with the cache cleared on disconnect and on `SET_SCALE` so both can re-assert. An installer flipping the project scale in Composer no longer has to wait for the device to drop.

**`esphome_bluetooth_coordinator` missed the docs sweep.** Added. Worded against its actual status values, since its `Connected` tracks the coordinator running rather than a device link.

**Changelog claim too narrow.** Reworded to "Dimming from the Control4 app was never affected", which is the user-facing fact, without implying the static list has no other readers.

**Test count.** 311 across eight files. My 146 was wrong.

## Two answers rather than changes

**Was the five-key flip minimal?** Honestly: `dimmer`, `set_level` and `ramp_level` were measured as a set, not individually, because the A/B variant flipped all three together. `supports_target` and `supports_brightness_stop` came from one reading, the brightness Target/Stop control being present and greyed on the static-False light, plus both precedents declaring `supports_target` True. So no, I did not isolate which of the five the test panel needs individually.

But the reason you give for wanting it minimal does not hold, and that is worth stating because it also affects DRV-113. `lightIsDimmer` ORs all three of the keys it reads:

```lua
toboolean(getLightSetup(deviceId, "dimmer"))
  or toboolean(getLightSetup(deviceId, "set_level"))
  or toboolean(getLightCapability(deviceId, "dimmer"))
  or toboolean(getLightCapability(deviceId, "set_level"))
  or lightHasRamp(deviceId)          -- itself ORs static ramp_level
```

Any one of `dimmer`, `set_level`, `ramp_level` being statically True is sufficient to make a mono light read as dimmable there. Narrowing three keys to one removes no exposure at all. `supports_target` and `supports_brightness_stop` are free, as you say, since that consumer never reads them. So minimality buys nothing here. Happy to run the isolation anyway if you want the documentation value, but it needs another Composer pass from a human and it will not change the DRV-113 surface.

**The precedents.** I did read both `driver.xml` files rather than citing them second hand, via sourcebot, so they are checkable without installing anything:

- `drivercentral.io/matter_light` - `<dimmer>True</dimmer> <set_level>True</set_level> <supports_target>True</supports_target>`, with `<supports_color>False</supports_color> <supports_color_correlated_temperature>False</supports_color_correlated_temperature> <supports_color_stop>False</supports_color_stop>`, rate bounds populated, and the same three BUTTON_LINK outputs on 300/301/302.
- `chowmain.software/chowmain_esphome_light` - `<supports_target>true</supports_target> <set_level>true</set_level> <ramp_level>true</ramp_level>` with `<dimmer>false</dimmer>`, the colour capabilities absent entirely, and `<brightness_rate_default>0</brightness_rate_default> <color_rate_default>0</color_rate_default>`.

Note Chowmain declares `dimmer=false` but `set_level`/`ramp_level` true, which by the OR above still trips DRV-113. It also ships a "Light Type" property (Auto Detect / On-Off Only / Dimmable / Dimmable CCT / Color / Color+CCT) as an installer override, which is a reasonable escape hatch if auto-detection ever misfires.

## Agreed and not acted on

**DRV-113 becomes a release prerequisite, and the failure flips direction.** Agreed on both, including that "a mono light publishes with a Brightness slider that does nothing" is the better failure than "a dimmable light publishes as a Switch". Nothing to change here; it is an ordering constraint on the release rather than on this PR, and DRV-113 already carries the enumeration-freeze half.

**Round 1's confirmation of the two-list reading, retracted in round 2.** For the record the retraction is right, and I made the same mistake first. The string table shows two capability name tables but nothing in the strings says which is which; a deployed A/B against `GET_SETUP` is the instrument that settles it.

## Still not covered

The Composer-side behaviour in this PR is measured, but by a human reading the UI, not by anything either of us can re-run. The four-way matrix, the colour controls staying usable with the colour tier False, and the app slider working throughout all rest on that.

derek-miller:
Thanks for the approval, and for retracting round 3 unprompted. I have to hand one of those back though: **the round 4 retraction is itself wrong, and round 3 was right.** `C4:GetDeviceData(<proxy id>)` returns the driver's declared capabilities, not a `light_v2.c4i` constant. Pushed `f3219ab` for finding 1.

## The retraction does not hold

Your instrument could not discriminate. `esphome_light` and `zigbee3_light` are the same code lineage, and zigbee3 adopted this PR's exact tier split in `beb7275` a few hours ago, so their capability blocks are genuinely identical right now. I confirmed that directly, comparing the two proxies' `<capabilities>` blocks:

```
esphome proxy 636 caps len=1984
zigbee  proxy 608 caps len=1984
identical=true
```

So "byte-identical across two drivers, therefore a constant" reads the same on both hypotheses. Two drivers that agree cannot tell them apart.

Three measurements that do.

**1. The proxy blob carries keys `light_v2` does not have.** Dumping the capability key set off proxy 636:

```
PROXY 636   #keys=45      DRIVER 635  #keys=45
   color_rate_default       true         true
   brightness_rate_default  true         true
   zero_ten_volt            false        false
   has_keypad               false        false
   number_of_loads          false        false
   min_max                  false        false
```

`color_rate_default` and `brightness_rate_default` are in the proxy's blob. Neither is a `light_v2` capability, which is the finding you confirmed yourself in round 1 by token search, and which a fresh install still seeding 750 ms independently supports. The proxy has no such concept to baseline from. They can only have come from a driver.xml.

**2. The converse.** `zero_ten_volt`, `has_keypad`, `number_of_loads` and `min_max` are all real `light_v2` static capabilities, from the same region you enumerated. They are **absent** from the proxy blob. A `light_v2.c4i` baseline would carry them. Our driver.xml does not declare them, and neither does the blob.

**3. A driver that actually differs.** I built a probe c4z from this branch's build, changed only two inert values and the name, and installed it without binding anything:

```
PROBE  proxy 649: hold_rate_max=86399999  cct_max=6499
SHIP   proxy 636: hold_rate_max=86400000  cct_max=6500
ZIGBEE proxy 608: hold_rate_max=86400000  cct_max=6500
```

`86399999` and `6499` exist in exactly one driver.xml on that controller and nowhere else. The proxy item reports them. That is conclusive. Probe driver has been deleted.

Worth noting your round 3 probe hit the same trap from the other side: "all three ESPHome test lights answered identically and device E came back `supports_color=False` despite being the RGBCT one". All three are the same driver, and `supports_color=False` is that driver's static declaration, so identical answers are what driver-derived predicts too. The RGBCT discrepancy is exactly right: the static list is not the merged state, which is why `GET_SETUP` exists.

### Consequences

- DRV-113 defect 1 is **not** pre-existing and universal. `getLightCapability` resolves to the bound driver's declared capabilities.
- This PR **does** change what Home Connect sees, in the direction you described in round 3.
- **DRV-113 goes back to being a prerequisite for this release.** Please treat the round 4 withdrawal as withdrawn. I would rather re-impose an ordering constraint than ship on a retraction that does not survive a third instrument.

I have recorded the three measurements on DRV-113 so this is not re-litigated from the string table a fourth time.

## Finding 1: accepted, and you were right twice over

Both halves. Nothing asserts the result, and `"FAHRENHEIT"` could not discriminate even with an assertion, because `getDisplayScale()` ends `or "F"` so a broken `normalizeScale` yields the same answer as a working one. I picked the value for the right reason and then picked one that cannot demonstrate it.

`f3219ab` makes the shim mutable and Celsius:

```lua
C4.TemperatureScale = "CELSIUS"
function C4:GetTemperatureScale()
  return C4.TemperatureScale
end
```

Verified the existing harness is unaffected: merged into `refs/pull/99/head`, 135 passed / 0 failed, so your five tests can go on top without adjusting anything.

**Yes please, open them against #99.** The tests are useless without `test_climate_driver.lua` and the mid-connection flip case is the one covering behaviour this branch added and I left unproven. Your counterfactuals are the part I care about most: 139/4 with `normalizeScale` broken and 141/2 with the re-check arm reverted is what makes them worth having.

## Minor, both fair

`supports_brightness_stop=True` genuinely has no precedent behind it. Matter declares it `False`, Chowmain omits it, and it went in from one Composer reading of a greyed Target/Stop control plus symmetry with `supports_target`. The driver.xml comment lists what each precedent declares without claiming that key; the commit message generalized past the evidence. It stays on the test-panel rationale alone.

The "cheap no-op" comment is fixed in `f3219ab`. The `SendToProxy` is skipped, but a `persist:get` and a `C4:GetTemperatureScale()` still run per state update, and the comment now says so and says what that buys.

derek-miller:
Fixed in `0099ffa`. Verified your finding from the source rather than taking the counts, and you are right on both halves.

`Persist:_get` caches into `self._persist[key]`, including the `EMPTY` sentinel when nothing is stored, and `require` makes it a module singleton, so `PersistGetValue` runs once per key for the driver's lifetime. `Persist:set` writes that cache on the way through, so `SET_SCALE` stays coherent without a re-read. And `getDisplayScale()` is an `or` chain, so an override short-circuits `C4:GetTemperatureScale()` entirely. One director read per update, none once an override is set.

I removed the note rather than correcting it to "one". The cost is not interesting at thermostat update rates, the function's own doc already says why it runs on every update, and this particular comment has now been wrong twice. That seems like the comment earning its removal.

Two other notes.

**On the independence failure.** Your diagnosis is sharper than mine and worth keeping. I had it as "the instrument could not discriminate", which is true but incomplete. The design was sound: two drivers behind one proxy is exactly right, because same-driver devices cannot separate a per-driver value from a constant. What broke it is that `zigbee3_light` adopted this PR's tier split in `beb7275` a few hours before you sampled, so the two drivers were identical at that moment. Independence is a property of the sample, not of the design, and it has to be re-checked at sample time. I would not have named that precisely, and it is the transferable part.

Your key-set equality run is also a better instrument than my probe driver, because it compares against the repo's own driver.xml rather than against another driver, so nothing about the rest of the project can interfere with it. Exact equality in both directions is not something a baseline can produce by coincidence.

**On the coverage gap.** Agreed it is still open and agreed the fixture was only the necessary half. Please do open the five against #99. The mid-connection flip is the one I most want, since it covers behaviour this branch added and I have no way to prove from here, and your counterfactuals are what make the set worth having rather than just green.

krishy2:
#101
I meant to display in Navigator if the device is connected.
One solution could be: 'MESSAGE_CHANGED ()' - that will appear on the front tab of a Thermostat proxy.
Other solution, which I mentioned earlier: <has_extras> capability, then an "extras" tab appears on Thermostat proxy. There could be a read-only parameter if its connected or not.

Just noticed there is a <has_connection_status> capability.

Since now there is a bool in programming, thats just an idea, not a request.

derek-miller:
Now that the driver drives the thermostat proxy correctly, Navigator shows the
thermostat differently on its own when it is offline, so you may not need to
build anything for this.

If you still want it in Programming, two variables:

- `IS_CONNECTED`. It was always in the list but never went False, because the
  driver had been sending a connection notification the thermostat proxy
  ignores. That is what this PR fixes, so it now goes False when the device
  drops.
- `Connected`, on the ESPHome driver and on every sub-driver.

Both land in the next release. `MESSAGE` is in that list too, but the driver does
not write to it, and I am not adding an Extras tab for now.

PR #105, opened 2026-09-06
fix(bluetooth): restart the BLE scanner instead of rebooting the device

derek-miller:
## Summary

BLE scanner recovery pressed a button it found by matching `"restart"` against the entity name. It now restarts the scanner itself over the API and presses nothing.

The original goal was to fix that name match. Investigating ESPHome's source turned up a better mechanism, so the detection code this started as is deleted rather than repaired.

## Why the name match could not just be tightened

ESPHome puts no platform information on the wire. `ListEntitiesButtonResponse` carries only `object_id`, `key`, `name`, `icon`, `disabled_by_default`, `entity_category`, `device_class` and `device_id`.

`device_class: restart` is not specific to the node restart button. In 2026.7.2 it is also declared by `factory_reset`, `safe_mode`, and the restart/reset buttons of `ld2410`, `ld2412`, `ld2420`, `ld2450`, `seeed_mr24hpc1` and `seeed_mr60fda2`. Read back off the test proxy:

```
Restart          device_class=restart  entity_category=1  icon=mdi:restart
Safe Mode Boot   device_class=restart  entity_category=1  icon=mdi:restart-alert
Factory Reset    device_class=restart  entity_category=1  icon=mdi:restart-alert
```

That is a stock `generic-bluetooth-proxy`. Same class, same category — only the icon differs, and a user can override it. Adding icon and category still leaves `ld2420`'s `revert_config`, which declares `device_class=restart`, `entity_category=config` and `icon=mdi:restart` and is indistinguishable from a real restart button. Only the name separates those two, and requiring the name to contain `"restart"` excludes every localized button (Neustart, Redémarrer, Riavvia, Reiniciar) — the original bug.

Guessing wrong is unattended and clears WiFi credentials, needing physical reflashing.

## What it does instead

`BluetoothScannerSetModeRequest`, already wired up in the client. ESPHome ignores a set-mode request for the mode the scanner is already in, so the round trip through the opposite mode is what makes it act — each accepted change stops the scan and re-arms continuous scanning, which the firmware relaunches once the stack is idle:

```cpp
void BluetoothProxy::bluetooth_scanner_set_mode(bool active) {
  if (this->parent_->get_scan_active() == active) return;
  this->parent_->set_scan_active(active);
  this->parent_->stop_scan();
  this->parent_->set_scan_continuous(true);
}
```

The mode is restored afterwards because active scanning is what BTHome devices need. No entity required, nothing pressed, and existing Bluetooth connections survive.

Recovery makes two attempts and then leaves the scanner alone. ESPHome already reboots itself for the failures it can detect: 255 failed scan starts (`handle_scanner_failure_`), or a scan that never terminates within twice the scan duration.

## Also fixed

`_startScannerWatchdog` returned early when no restart button was found, so proxies without one had no scanner monitoring at all. Since the button was found by name, that included every proxy whose restart button is named in another language. It now starts on any proxy that reports scanner state.

It is gated on the `SCANNER_STATE` (`0x40`) feature flag, which was declared but never read. Without it the cached scanner state never leaves its `IDLE`/uninitialized default, so the watchdog could only ever take its ignore branch.

## Test plan

- [x] `make test` — 326 assertions, 15 new covering the escalation ladder, including one asserting nothing but scanner mode is ever touched
- [x] `make check`, `make build-nodocs`
- [x] `make build` (not just `build-nodocs`, which skips the `README.md` generation the dirty-tree check needs)
- [x] Real ESP32 proxy on ESPHome 2026.7.2: two in-place restarts, each a clean `RUNNING → STOPPING → IDLE → STARTING → RUNNING` cycle with the mode restored, advertisements flowing throughout, device left in `RUNNING (ACTIVE)`
- [x] Past the budget: one warning, watchdog keeps watching, nothing pressed

**What the hardware run does not establish:** it exercises the mechanism against a *healthy* scanner, which is not the state the code runs in. That the set-mode round trip drives the state machine is proven; that it recovers a *wedged* scanner is not, and would need a reproduction of the wedge. Those come apart if the stall is in the ESP-IDF stack rather than in ESPHome's own state.

DRV-115 (To be discussed) tracks the remaining gap. Framed precisely: a scanner that cycles normally but returns no advertisements never fails a scan start and never exceeds the scan timeout, so neither ESPHome reboot path fires. That is the exact state this watchdog exists to detect, and after two in-place restarts nothing else happens. The ticket is about whether that is acceptable, not about restoring something ESPHome already covers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,13 +37,15 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
   lights that do support brightness. Dimming from the Control4 app was never
   affected.
-- Fixed Bluetooth proxy scanner recovery rebooting the device, which it did by
-  pressing whichever button had "restart" in its name. On some devices that is a
-  factory reset button. Recovery now restarts the scanner itself, which is
-  quicker, keeps existing Bluetooth connections alive, and never presses
-  anything. (DRV-115)
-- Fixed Bluetooth proxies never watching for a stuck scanner unless the device
-  exposed a restart button. Scanner recovery now runs on every Bluetooth proxy.
+- Fixed Bluetooth proxy scanner recovery never running on many devices. It
+  needed a button with "restart" in its name, so proxies whose restart button is
+  named in another language, or that expose no such button, silently had no
+  recovery at all. Recovery no longer needs a button and runs on any proxy that
+  reports its scanner state.
+- Fixed Bluetooth proxy scanner recovery rebooting the device, and choosing what
+  to press by name, which could press an unrelated button. It now restarts the
+  scanner itself, which is quicker and keeps existing Bluetooth connections
+  alive. (DRV-115)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,13 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
   lights that do support brightness. Dimming from the Control4 app was never
   affected.
+- Fixed Bluetooth proxy scanner recovery rebooting the device, which it did by
+  pressing whichever button had "restart" in its name. On some devices that is a
+  factory reset button. Recovery now restarts the scanner itself, which is
+  quicker, keeps existing Bluetooth connections alive, and never presses
+  anything. (DRV-115)
+- Fixed Bluetooth proxies never watching for a stuck scanner unless the device
+  exposed a restart button. Scanner recovery now runs on every Bluetooth proxy.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -928,6 +928,15 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
   lights that do support brightness. Dimming from the Control4 app was never
   affected.
+- Fixed Bluetooth proxy scanner recovery never running on many devices. It
+  needed a button with "restart" in its name, so proxies whose restart button is
+  named in another language, or that expose no such button, silently had no
+  recovery at all. Recovery no longer needs a button and runs on any proxy that
+  reports its scanner state.
+- Fixed Bluetooth proxy scanner recovery rebooting the device, and choosing what
+  to press by name, which could press an unrelated button. It now restarts the
+  scanner itself, which is quicker and keeps existing Bluetooth connections
+  alive. (DRV-115)
 
 ### Changed
 
--- a/drivers/esphome/driver.lua
+++ b/drivers/esphome/driver.lua
@@ -592,15 +592,6 @@ function RefreshStatus()
           else
             log:debug("No Entities['%s']:discovered() handler", entity.entity_type)
           end
-
-          -- Detect restart button for scanner recovery
-          if entity.entity_type == "button" then
-            local buttonName = entity.name or ""
-            if buttonName:lower():find("restart") then
-              log:info("Found restart button entity: %s", ESPHomeClient.describeEntity(entity))
-              bluetoothProxyCapability:setRestartButtonKey(entity.key)
-            end
-          end
         end
 
         return entities
--- a/src/esphome/capabilities/bluetooth_proxy.lua
+++ b/src/esphome/capabilities/bluetooth_proxy.lua
@@ -190,6 +190,14 @@ function BluetoothProxyCapability:_startScannerWatchdog()
     return
   end
 
+  -- The watchdog reads scanner state to tell a stalled scanner from one that
+  -- is legitimately stopped. Without that flag the cached state never leaves
+  -- its default and every check would be a no-op.
+  if bit32.band(self._featureFlags, FEATURE_FLAGS.SCANNER_STATE) == 0 then
+    log:debug("Scanner watchdog not started: device does not report scanner state")
+    return
+  end
+
   log:debug("Starting scanner watchdog (interval: %ds)", SCANNER_WATCHDOG_TIMEOUT_SECONDS)
   self._scannerWatchdogActive = true
   self._scannerWatchdogSeen = false
--- a/src/esphome/capabilities/bluetooth_proxy.lua
+++ b/src/esphome/capabilities/bluetooth_proxy.lua
@@ -30,6 +30,18 @@ local SCANNER_WATCHDOG_TIMEOUT_SECONDS = 90
 --- Timer key for the scanner watchdog
 local SCANNER_WATCHDOG_TIMER_KEY = "BLEScannerWatchdog"
 
+--- How many times to restart the scanner in place before leaving it to
+--- ESPHome's own recovery. Each attempt costs one watchdog interval.
+local SCANNER_RECOVERY_RESTART_ATTEMPTS = 2
+
+--- Timer key for restoring the scanner mode after an in-place restart.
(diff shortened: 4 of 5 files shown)


derek-miller (commented):
Verified the reasoning rather than the diff, since the conclusion overturns the brief this task started from, and that brief was mine.

## My suggested fix was worse than the bug

I told this task to gate on `device_class == "restart"`. Checked against ESPHome 2026.7.2:

```
factory_reset  device_class=DEVICE_CLASS_RESTART  entity_category=ENTITY_CATEGORY_CONFIG  icon=ICON_RESTART_ALERT
safe_mode      device_class=DEVICE_CLASS_RESTART  entity_category=ENTITY_CATEGORY_CONFIG  icon=ICON_RESTART_ALERT
restart        device_class=DEVICE_CLASS_RESTART  entity_category=ENTITY_CATEGORY_CONFIG  icon=ICON_RESTART
```

Ten button platforms declare that class. So the gate I proposed would select `Factory Reset` on any device exposing one, and press it unattended.

The part I got backwards: `"Factory Reset"` does not contain `"restart"`, so the name match this was meant to fix would never have pressed it. My replacement would have. I moved the failure from "misses a localized button" to "wipes WiFi credentials and needs physical reflashing", and called it the safer option.

`ld2420` closes the door on tightening it further, exactly as the PR says:

```
apply_config   device_class=RESTART  entity_category=CONFIG  icon=ICON_RESTART_ALERT
revert_config  device_class=RESTART  entity_category=CONFIG  icon=ICON_RESTART
```

`revert_config` carries the identical triple to the real restart button. Class, category and icon together still cannot separate them, and only the name can, which is the original bug. Deleting the detection rather than repairing it is the right call.

## The replacement mechanism checks out

`bluetooth_proxy.cpp` in 2026.7.2 is as quoted, including the early return that makes the round trip necessary:

```cpp
void BluetoothProxy::bluetooth_scanner_set_mode(bool active) {
  if (this->parent_->get_scan_active() == active) {
    return;
  }
  ESP_LOGD(TAG, "Setting scanner mode to %s", active ? "active" : "passive");
  this->parent_->set_scan_active(active);
  this->parent_->stop_scan();
  this->parent_->set_scan_continuous(true);
```

The trailing comment in the source, "Set this to true to automatically start scanning again when it has cleaned up", is the thing that makes this a restart rather than a stop. Needing no entity at all is a strictly better position than any name or metadata heuristic.

## The watchdog fix is real

On `main`, `_startScannerWatchdog` returns at line 185 when `_restartButtonKey` is unset, so a proxy without a restart button had no scanner monitoring whatsoever. That is a separate bug from the one this task was filed for and worth calling out in its own right, since it silently disabled the feature on some hardware rather than misbehaving visibly.

## Not verified

The ESP32 hardware run. I have no BLE proxy on the test controller, so the `RUNNING -> STOPPING -> IDLE -> STARTING -> RUNNING` cycle and the mode restoration are yours alone.

Deferring the reboot-fallback question to DRV-115 is right. Reintroducing a reboot would put back the need to identify a restart button, which is the whole problem.

derek-miller:
Pushed `5a40d63`. Both reviews were right on every point I could check; taking them in turn.

## Blocker: dirty tree — fixed

`README.md` is generated from `CHANGELOG.md` and I never regenerated it, because my test plan ran `make build-nodocs`, which skips `docs-readme`. Exactly the case a changelog-only edit cannot be caught by. Ran the full `make build` and committed the result; `make docs-readme` is now idempotent against the tree.

## 1. Watchdog gating — fixed, and the claim was wrong

Confirmed: `FEATURE_FLAGS.SCANNER_STATE` was declared at `bluetooth_proxy.lua:22` and read nowhere. Without the flag, `_btScannerState` keeps its constructor default (`state = IDLE`, `initialized = false`, `client.lua:192`) and is reset to the same on every reconnect (`:509`), so the timer would fire every 90 seconds forever and always take the ignore branch.

`_startScannerWatchdog` now returns early unless the flag is set, and there is a test for both directions. The changelog no longer says "every Bluetooth proxy" — it says "any proxy that reports its scanner state".

## 4. The factory reset claim — withdrawn

You are right, and I want to be precise about the error since it is the second time this has bitten this task. The shipped code matched `"restart"` as a substring of the entity name. `"Factory Reset"` does not contain it. So the factory reset hazard belonged to the `device_class` gate I rejected, not to the code being replaced — I imported the danger from the rejected design into the description of the shipped one.

Restated on what the evidence actually shows: the name match silently armed no recovery at all whenever no button name contained "restart", which is every localized restart button. That is the user-visible bug, and it is the one the readback demonstrates.

## 5. Restore leg — covered

Correct that `client.calls[2] == nil` passed only because the 5s timer had not fired. Added `ShimFireTimers()` and an assertion that the mode goes back to active, plus the same in the bounded-attempts test so its call count now covers both legs. 19 assertions in that file, 330 across the suite.

## 2 and 3 — wording corrected, no code change

Both fair, and both were overclaims rather than errors.

DRV-115 is retitled *"One BLE scanner failure mode has no recovery: cycling but deaf"* and reframed on your terms: neither ESPHome reboot path fires for a scanner that cycles normally and returns nothing, because `scan_start_time_` resets each cycle and no start fails. That is the state this watchdog exists for. The ticket now presents it as a gap nothing covers, not a fallback ESPHome mostly covers, and lists surfacing it in the status property or an installer-selected button as the options that avoid re-introducing identification.

On the hardware run: agreed the checkbox read as more than it proved. The PR body now says plainly that it exercises the mechanism against a healthy scanner, and that recovering a wedged one is untested and would need a reproduction of the wedge.

## On the first review

Noting for the record that the `device_class` gate the brief asked for would have selected `Factory Reset` on the very device used for testing — all three of its buttons report `device_class=restart` with `entity_category=config`. That is why this ended up deleting the detection instead of fixing it.

`pressButton` is indeed still used by `entities/button.lua`; nothing was orphaned.

derek-miller (commented):
`5a40d63` is a clear improvement on both counts. Two verifications and one thing worth knowing.

## The new guard excludes nobody on current firmware

`SCANNER_STATE = 0x40` matches ESPHome's `FEATURE_STATE_AND_MODE = 1 << 6`. More usefully, that flag is set unconditionally:

```cpp
uint32_t get_feature_flags() const {
  uint32_t flags = 0;
  flags |= BluetoothProxyFeature::FEATURE_PASSIVE_SCAN;
  flags |= BluetoothProxyFeature::FEATURE_RAW_ADVERTISEMENTS;
  flags |= BluetoothProxyFeature::FEATURE_STATE_AND_MODE;
  if (this->active_) { ... }
```

It sits outside the `active_` block, so every proxy on 2026.7.2 reports it, passive ones included. The guard is right to exist, and in practice it narrows nothing on current ESPHome. Only firmware predating the flag falls outside it, which is the correct place for the line to sit and is a much smaller exclusion than the restart-button one it replaces.

Worth having anyway. Your reasoning holds regardless of how many devices it catches: without scanner state the cached value never leaves its default and every check is a no-op, so a watchdog that ran there would be theatre.

## The restore assertion is the important test change

The previous version checked the flip to passive and that nothing else was touched, then stopped. It never asserted the mode came back. A restore that silently failed would have left proxies stuck in passive mode, which is precisely what BTHome devices cannot work with, and the old test would have stayed green through it. Driving `ShimFireTimers()` and asserting `mode:active` closes that, and extending the budget case to four calls rather than two keeps the escalation ladder honest.

330 passed / 0 failed on this head, up from 326.

## The changelog rewording is more accurate

Leading with "never running on many devices" is the right order: that is the bug a user experienced, and the reboot behaviour is what it did on the minority of devices where it ran at all. The second entry no longer claims recovery runs on every proxy, which the new guard would have made untrue.

Nothing outstanding from me.

PR #108, opened 2026-09-08
fix: tell sub-drivers when the ESPHome device goes away

derek-miller:
Two paths left a sub-driver reporting connected against a device that was not there. Found while putting a thermostat into a disconnected state to look at Navigator, so both are reproduced rather than inferred.

## 1. Changing a connection setting reset the transition state

`wasConnected` lived inside `Connect()`, which `OPC.IP_Address`, `OPC.Port`, `OPC.Password`, `OPC.Encryption_Key` and `OPC.Use_OpenSSL` all call. Each call built a fresh closure with `wasConnected = false` plus a new heartbeat timer, so the heartbeat that followed saw no true to false edge and `notifyEntitiesDisconnected()` never ran.

The heartbeat guards on the transition:

```lua
local connected = esphome:isConnected()
if wasConnected and not connected then
  log:info("Connection lost, notifying sub-drivers")
  notifyEntitiesDisconnected()
end
wasConnected = connected
```

Measured on a CORE 1 with a host-mode climate mock. Repointing the parent at an unreachable address:

```
before   parent Connected=0, driver status=Connection failed: Connection refused
         climate sub-driver status=Connected, proxy IS_CONNECTED=1
```

It stayed that way for over three minutes while the parent logged `Connection refused` every ten seconds, and no `Connection lost, notifying sub-drivers` line ever appeared. I had to send `UPDATE_DISCONNECT` by hand to clear it.

After hoisting `wasConnected` to file scope, the identical repro:

```
18:02:14 [INFO ]: Connection lost, notifying sub-drivers
18:02:14 [DEBUG]: Connecting to ESPHome device at <unreachable>
after     parent=0  driver=Disconnected  Connected=0  IS_CONNECTED=0
```

No manual intervention.

## 2. Water heaters were never told at all

`WaterHeaterEntity.TYPE` is `WATER_HEATER`, but they share the ESPHome Climate sub-driver, and `ClimateEntity:disconnected()` only walks the `CLIMATE` namespace:

```lua
for _, binding in pairs(bindings:getDynamicBindings(self.TYPE)) do
```

So no path reached a water heater, including an ordinary device drop. Added `WaterHeaterEntity:disconnected()`.

Checked the other entity types while I was there. Of the five with proxy sub-drivers, `climate`, `fan`, `light` and `lock` already implement `disconnected()`; `water_heater` was the only gap. The rest (`binary_sensor`, `switch`, `cover`, `sensor`, `number`, `select`, `text`, `text_sensor`, `date`, `time`, `datetime`, `event`, `button`) are variable, relay and contact backed rather than proxy sub-drivers, so they have no proxy connection state to go stale. Left alone.

## Changelog

The unreleased entry from #104 already reads "Fixed thermostats and water heaters staying shown as connected after the ESPHome device went offline." That was accurate about thermostats and not about water heaters, since the handler did not exist. It becomes true with this change, so it is left as-is rather than duplicated.

The connection setting bug is separate and pre-existing, so it gets its own entry. Before #104 it was masked on thermostats, because `ONLINE_CHANGED` is a no-op on `thermostatV2`, but lights, fans and locks have been affected all along.

## Test plan

- [x] `make test` 429 passed, 0 failed
- [x] `make check`, `make build`
- [x] Repointing a connected parent at an unreachable address now drives the sub-driver to Disconnected and `IS_CONNECTED` to 0, verified on a CORE 1 (fails without the change, on the same rig)
- [x] `Connection lost, notifying sub-drivers` appears in the parent log where it previously never did
- [ ] Water heater disconnect not exercised on hardware. There is no water heater entity on the test rig, so that half is by inspection: same binding walk as the climate handler, one namespace over

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,9 +37,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
   lights that do support brightness. Dimming from the Control4 app was never
   affected.
-- Fixed lights, thermostats, fans and locks still showing as connected after the
-  ESPHome driver's IP address, port or credentials were changed. They now go
-  offline with the device until it reconnects.
+- Fixed lights, thermostats, water heaters, fans and locks still showing as
+  connected after the ESPHome driver's IP address, port or credentials were
+  changed or cleared. They now go offline with the device until it reconnects.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
   lights that do support brightness. Dimming from the Control4 app was never
   affected.
+- Fixed lights, thermostats, fans and locks still showing as connected after the
+  ESPHome driver's IP address, port or credentials were changed. They now go
+  offline with the device until it reconnects.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -928,6 +928,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
   lights that do support brightness. Dimming from the Control4 app was never
   affected.
+- Fixed lights, thermostats, fans and locks still showing as connected after the
+  ESPHome driver's IP address, port or credentials were changed. They now go
+  offline with the device until it reconnects.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -928,9 +928,9 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 - Fixed Composer's test panel greying out the dimming controls for ESPHome
   lights that do support brightness. Dimming from the Control4 app was never
   affected.
-- Fixed lights, thermostats, fans and locks still showing as connected after the
-  ESPHome driver's IP address, port or credentials were changed. They now go
-  offline with the device until it reconnects.
+- Fixed lights, thermostats, water heaters, fans and locks still showing as
+  connected after the ESPHome driver's IP address, port or credentials were
+  changed or cleared. They now go offline with the device until it reconnects.
 
 ### Changed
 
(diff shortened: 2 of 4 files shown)


derek-miller:
Both findings were right and both are fixed in `b75ad63`. Thanks for checking the
namespace claim structurally, since the hardware repro genuinely could not see it.

## The residual path is closed

You are right that the two guards return before the transition check and both
drop the connection on the way out, and that the `isConfigured` one also cancels
the heartbeat so nothing later corrects it. Both now go through a small helper
that disconnects and notifies when the driver was previously connected, rather
than moving the transition check upwards. Moving it up would not have worked:
on the cycle where a guard fires, `isConnected()` is still true at the top of
the heartbeat, so the edge would be missed and there is no next cycle.

Verified on a CORE 1, blanking `IP Address` on a connected device:

```
18:20:13 [INFO ]: Connection closed, notifying sub-drivers
BLANK IP: parent=0  driver=Disconnected  Connected=0  IS_CONNECTED=0
```

Your point about the licence guard being live in a shipped build is what moved
this from a follow-up to part of this PR. Same shape, and a lapsed licence would
have stranded every sub-driver online.

## Changelog

Water heaters added, and "changed" widened to "changed or cleared" so the entry
stops over-claiming the case you found.

## On the unexercised half

Fair, and your read is right that the only uncovered link is the namespace walk,
since the `UPDATE_DISCONNECT` lands on the same already water-heater-aware
handler my thermostat repro drove. Noted that `test/dummy_water_heater.yaml` is
runnable on the same rig. I have not closed it by measurement because the mock
the rig was pointing at has since gone away, so I would rather leave the caveat
honest than claim a run I did not do.

derek-miller:
The correction is right and I have confirmed it rather than taking it.
`ESPHomeClient:setConfig` calls `self:disconnect()` unconditionally as its second
statement (`client.lua`), and `Connect()` calls `setConfig` at `driver.lua:437`,
well before it arms the timer at 545 and calls `heartbeat()` at 546. So on that
synchronous first beat the connection is already dropped and `isConnected()`
returns false, not true as I claimed.

So a hoist would have worked for both guards, and my stated reason for not doing
one was wrong. Worth having that on the record for the reason you gave: "the
transition check cannot see a config change" is exactly the kind of premise that
gets reused, and it is false.

The code comment on the helper does not carry the bad claim, so nothing to
change there. Sticking with the helper for the reasons you list, notifying on the
same synchronous call rather than up to ten seconds later and not depending on
the heartbeat surviving, rather than because a hoist could not work.

Also useful: checking the other five `esphome:disconnect()` call sites and both
distributions preprocessing cleanly are the two things I did not do. Good to know
the OSS build leaves no orphaned local.

Issue #55, opened 2026-04-18
Possible issue with bool variables in ESPhome driver


dwaynenewsome:
I have been trying to track down an issue with bools not changing in the variables agent in C4 with my air freshner "driver".  I reproduced the issue with the ratdgo device.  I think this is a bug.    I changed the LED state in Ratdgo Esphome page.   I had programming on the variable change in c4.  It fires and gets true/false correctly.  However, the variable never changes to true it is always false.   The time stamp updates, but the value does not change to true.  If you need any logs or anything else i can send.

derek-miller:
Anything you can share to recreate the issue?

dwaynenewsome:
Using RATGDO to set the LED to true <img width="1197" height="398" alt="Image" src="https://github.com/user-attachments/assets/ad454518-fa01-4d3d-8293-3d99e2fe9784" />  The variable timestamp changes, so it gets the call, but stays false 

<img width="1290" height="216" alt="Image" src="https://github.com/user-attachments/assets/59df9cc3-5086-4dcb-9b12-2c33ad71902d" />  This programming event fires and it is TRUE  

<img width="964" height="381" alt="Image" src="https://github.com/user-attachments/assets/5b78f9d7-faef-427b-88f2-bcfc883b0a2f" />

PR #110, opened 2026-09-09
feat: let event entities drive loads as keypad buttons

### LINKED ISSUE #103: BUTTON object direction and feedback

derek-miller:
Closes #103.

An ESPHome `event:` entity previously created only Control4 events and a
`Last Event` variable, so a touch button on the device could reach a light
only through Programming. Each declared event type now also gets its own
`BUTTON_LINK` connection on the keypad side, bindable straight to a load in
Composer.

`button:` entities are unchanged. Their connection points the other way and
that stays correct: ESPHome has no message for a device-side button press, so
a `button:` can only ever be pressed by the controller.

One event sends one `DO_CLICK`. `DO_CLICK`, `DO_PUSH` and `DO_RELEASE` are
three phases of one press gesture, not three dialects to cover: a load that
implements all of them reads the triple as a click immediately followed by a
hold that ends where it began, and the click is undone. An ESPHome event is a
gesture the device has already completed, so there is no button-down to report.

## Verification

Run against a mock carrying both a dimmable light and an `event` entity on one
device, so a real event drives a real load through a real binding.

Discovery created one binding per declared type, all reported by the director as
`BUTTON_LINK` / **Input**, alongside a `button:` entity binding that stayed
**Output**:

```
bindingId 11  Test Button press          BUTTON_LINK  Input
bindingId 12  Test Button double_press   BUTTON_LINK  Input
bindingId 13  Test Button long_press     BUTTON_LINK  Input
bindingId 10  Test Action Button         BUTTON_LINK  Output
```

Measuring the light component's own state rather than the driver's, a Toggle
button link driven from the same live binding:

```
start        messages     final state              verdict
OFF          DO_CLICK     on=1 brightness=1.000    correct
OFF          triple       on=1 brightness=0.060    wrong, should be 1.000
ON  1.000    triple       on=1 brightness=0.990    wrong, should be off
ON  0.990    DO_CLICK     on=0 brightness=0.000    correct
```

The two stuck cases are the ramp getting a single step in before
`brightnessRampStopped()` freezes it.

End to end after the fix, a real repeated `event.trigger("press")` alternates
cleanly:

```
FIRING probe press -> on=1 brightness=1.000
FIRING probe press -> on=0 brightness=0.000
FIRING probe press -> on=1 brightness=1.000
```

`test/test_event_button_link.lua` asserts one `DO_CLICK` and nothing else, on
the binding for that event type. Reverting to the triple fails it.

## Note for the merger

`test/c4_shim.lua` is template-managed and this repo is pinned at `v0.9.21`, so
the stubs here are a local divergence until
finitelabs/control4-driver-template#55 lands and is tagged. Merge that first, so
the next copier update is a no-op on this file rather than a conflict.

The same self-cancelling triple is pre-existing in `esphome_bthome` and
`esphome_switchbot`; tracked in DRV-120, out of scope here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ## Unreleased
 
+### Added
+
+- Added a button connection for each event type an ESPHome event entity
+  declares, so a touch button or gesture on the device can drive a light, scene
+  or any other load directly, without Programming.
+
 ### Fixed
 
 - Fixed an automatic update sometimes leaving companion drivers on the previous
--- a/README.md
+++ b/README.md
@@ -487,15 +487,20 @@ the matching read-only Device Info properties.
 | Cover         | `CONTACT_SENSOR`                | Open/closed state contacts                                        |
 | Cover         | `RELAY`                         | Open/close/stop control relays                                    |
 | Button        | `BUTTON_LINK`                   | Allows other devices to trigger button                            |
+| Event         | `BUTTON_LINK`                   | One connection per event type, triggers other devices             |
 | Climate       | `ESPHOME_CLIMATE`               | Bind to ESPHome Climate sub-driver [\*\*](#climate-services-note) |
 | Fan           | `ESPHOME_FAN_N_SPEED[_REVERSE]` | Bind to ESPHome Fan sub-driver                                    |
 | Light         | `ESPHOME_LIGHT`                 | Bind to ESPHome Light sub-driver                                  |
 | Lock          | `ESPHOME_LOCK`                  | Bind to ESPHome Lock sub-driver                                   |
 | Water Heater  | `ESPHOME_CLIMATE`               | Bind to ESPHome Climate sub-driver                                |
 
-> **Note:** Sensor, Number, Select, Text, Text Sensor, Date, Time, Datetime, and
-> Event entities do not create bindings. They expose data only through variables
-> and events.
+> **Note:** Sensor, Number, Select, Text, Text Sensor, Date, Time and Datetime
+> entities do not create bindings. They expose data only through variables and
+> events.
+
+> **Note:** The Button and Event binding classes are the same but point opposite
+> ways. A Button binding lets another device press the ESPHome button; an Event
+> binding lets the ESPHome device act as a keypad button and drive a load.
 
 > **Note:** Water Heater entities share the `ESPHOME_CLIMATE` binding class and
 > are controlled via the ESPHome Climate sub-driver. Device modes (such as Eco,
@@ -536,8 +541,9 @@ bindings are created separately from the entity bindings above:
 
 > **Note:** Event entities are stateless triggers (button presses, gestures,
 > doorbell rings). Each discovered event type creates a Control4 event that can
-> be used in programming. The `{name} Last Event` variable tracks the most
-> recent event type.
+> be used in programming, plus a `BUTTON_LINK` connection named
+> `{name} {event_type}` that can be bound straight to a load. The
+> `{name} Last Event` variable tracks the most recent event type.
 
 ### Commands
 
--- a/README.md
+++ b/README.md
@@ -919,6 +919,12 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ## Unreleased
 
+### Added
+
+- Added a button connection for each event type an ESPHome event entity
+  declares, so a touch button or gesture on the device can drive a light, scene
+  or any other load directly, without Programming.
+
 ### Fixed
 
 - Fixed an automatic update sometimes leaving companion drivers on the previous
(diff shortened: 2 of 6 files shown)


derek-miller:
@svc-finitelabs please review

derek-miller:
Confirmed and fixed in 8dbb111. The finding is right, and I reproduced it on the
rig rather than taking the harness result on faith.

I rebuilt the mock so one device carried both a real dimmable light and an
`event` entity, which the earlier run did not have: device 634 was pointed at
the inputs mock, so the light bindings were stale and no load was actually
behind them. That is why the `RFP.DO_CLICK` counter looked clean. It counted a
call that arrived and then got undone.

End to end first, a real `event.trigger("press")` bound to the light's Toggle,
repeated:

```
FIRING probe press
mono_light state: on=1 brightness=0.004
mono_light state: on=1 brightness=1.000
mono_light state: on=1 brightness=0.060   <- settles here
```

The light never latched. Then the four cells, driven on the same live binding,
reading the light component's own state rather than the driver's:

```
start        messages     final state              verdict
OFF          DO_CLICK     on=1 brightness=1.000    correct
OFF          triple       on=1 brightness=0.060    wrong, should be 1.000
ON  1.000    triple       on=1 brightness=0.990    wrong, should be off
ON  0.990    DO_CLICK     on=0 brightness=0.000    correct
```

My numbers differ slightly from yours in the stuck cases, 0.060 and 0.990 rather
than settled 0 and 100, which is the ramp getting one step in before
`brightnessRampStopped()` freezes it. Same mechanism, same conclusion.

After the fix, the same repeated event alternates cleanly: 1.000, 0.000, 1.000.

## On the open question

You were right not to assert `DO_CLICK` alone without checking, so I checked.
Two things came back.

Against it: a load that acts on `DO_PUSH` and ignores `DO_CLICK` does exist in
the wild. `chowmain_elk_m1_v2` has a `BUTTON_TO_FUNCTION.DO_PUSH` that presses a
function key and a `DO_CLICK` that is an empty body.

For it: the dominant idiom among button *sources* is `DO_PUSH` then `DO_CLICK`,
with no `DO_RELEASE`. `chowmain_philips_hue_light` sends exactly that for single,
double and triple click and for short and long release, and the two-message form
also shows up in `chowmain_link_server_keypad` and `chowmain_bigassfans`. Our own
light driver is built for that sequence: `rewindToPrePress()` exists to undo the
hold ramp when a `RELEASE_CLICK` follows a `PRESS`. `DO_RELEASE` is the hold
ending, and it is the alternative to a click, not a suffix to one.

So `DO_PUSH` + `DO_CLICK` would also work against our light. I went with
`DO_CLICK` alone anyway, for two reasons. It is correct for every receiver in
this repo, where `esphome_fan` implements only `DO_CLICK` and
`esphome_switchbot` acts on `DO_CLICK` or `DO_PUSH` and would fire twice on the
pair. And a synthetic `DO_PUSH` claims a button-down that never happened; an
ESPHome event carries no press and release, only a completed gesture.

## Scope

Agreed it is pre-existing, and I confirmed the switchbot double-fire on top of
what you found, so I filed DRV-120 for both drivers rather than widening this PR.

## Test

Added `test/test_event_button_link.lua`. It asserts one `DO_CLICK` and nothing
else, on the binding for that event type. Reverting to the triple fails it:

```
FAIL press  -> got "10:DO_CLICK,10:DO_PUSH,10:DO_RELEASE", want "10:DO_CLICK"
```

It needed `C4:AddEvent` / `DeleteEvent` / `FireEventByID` in `test/c4_shim.lua`,
which were missing. Added here; they belong in the template, same as the
`GetTemperatureScale` gap, and I will roll them up separately.

derek-miller:
Correction on the shim: `test/c4_shim.lua` is template-managed, so that edit
should not have originated here. It is now up as
finitelabs/control4-driver-template#55, with the stubs given their own section
and covered in the template's own `test_c4_shim.lua`.

The copy in this branch stays for the moment so the new test runs, and comes out
on the copier update once #55 lands and is tagged. Worth sequencing that before
this merges rather than after, so the sweep is a no-op instead of a conflict.

derek-miller:
Both items handled.

PR body rewritten to describe what this head actually does: the emitted sequence
is one `DO_CLICK`, the four-cell table is the light component's own state, and
the `RFP.DO_CLICK` counter is gone, since as you say it is the measurement that
could not see the bug. The merge-time note about template sequencing and DRV-120
is in there too.

The `esphome_bthome` comment mismatch is real and worse than one line: the doc
comment at `:618` says "DO_PUSH followed by DO_CLICK", the code sends
`DO_CLICK`, `DO_PUSH`, `DO_RELEASE`, and the `log:debug` between them describes
a third order. Added to DRV-120 so whoever takes it settles all three together.

Your Elk framing is the right one to carry forward and I have nothing to add to
it: there is no message set that satisfies both that receiver and our SwitchBot,
so a third-party load going dead beats double-firing a device in this repo, and
if it ever matters the answer is per-consumer rather than a different global
triple.

Holding this until control4-driver-template#55 is merged and tagged.

PR #113, opened 2026-09-11
feat: let event entities drive loads as keypad buttons

### LINKED ISSUE #103: BUTTON object direction and feedback

derek-miller:
Closes #103. Re-opens the work from #110, which was merged without a go-ahead
and reverted in #112. Since then the branch has picked up template v0.9.22 (its
shim now carries the event stubs, so this branch has none) and `main` as of
#123.

An ESPHome `event:` entity previously created only Control4 events and a
`Last Event` variable, so a touch button on the device could reach a light
only through Programming. Each declared event type now also gets its own
`BUTTON_LINK` connection on the keypad side, which can be bound straight to a
load in Composer.

`button:` entities are unchanged. Their connection points the other way, which
stays correct: ESPHome has no message for a device-side button press, so a
`button:` can only ever be pressed by the controller.

## Protocol

An event sends `DO_PUSH` then `DO_CLICK`, the pair a Control4 keypad sends for a
tap. That matches what #121 settled for the BTHome and SwitchBot senders.

#110 sent `DO_CLICK` alone, to avoid double-firing the SwitchBot receiver and
because a push-then-click pair seemed unnecessary. Since #121:

- the SwitchBot receiver fires once per tap
- loads that act only on `DO_PUSH` (they exist) need the push to fire at all

`DO_RELEASE` is not sent, because it ends a hold rather than a click.

## Verification

Tested on a controller with a host-mode mock that carries a dimmable light and
an `event` entity on the same device.

Discovery creates one binding per declared type. The director reports each as
`BUTTON_LINK` / **Input**, while a `button:` entity's binding stays **Output**:

```
bindingId 11  Test Button press          BUTTON_LINK  Input
bindingId 12  Test Button double_press   BUTTON_LINK  Input
bindingId 13  Test Button long_press     BUTTON_LINK  Input
bindingId 10  Test Action Button         BUTTON_LINK  Output
```

End to end at this head: a device-side `event.trigger("press")` every 25s, with
the event bound to the light's Toggle button link. Each press toggles it cleanly,
read from the light component's own state:

```
08:26:50 FIRING probe press -> on=1 brightness=1.000
08:27:15 FIRING probe press -> on=0 brightness=0.000
08:27:40 FIRING probe press -> on=1 brightness=1.000
08:28:05 FIRING probe press -> on=0 brightness=0.000
```

`test/test_event_button_link.lua` asserts `DO_PUSH` then `DO_CLICK` and nothing
else, on the binding for the event type that fired. It fails for click only, for
the old triple, and for the pair in reverse order. `make check` is clean and all
suites are green.

`test_button_link_protocol.lua`'s "nothing sends DO_RELEASE" scan now also reads
`src/`, since `entities/event.lua` is the first sender outside `drivers/`. With
`DO_RELEASE` added to `event.lua` it now fails; before this change it passed.

## Known limitation

The binding key includes the event type. Renaming a type in YAML therefore
creates a new connection and leaves the old one in Composer, still bound but
dead. The Control4 event created just above behaves the same way, so this is
consistent, not new.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ Template for a new release entry (copy below the heading, fill in, uncomment):
 
 ## Unreleased
 
+### Added
+
+- Added a button connection for each event type an ESPHome event entity
+  declares, so a touch button or gesture on the device can drive a light, scene
+  or any other load directly, without Programming.
+
 ### Fixed
 
 - Fixed an automatic update sometimes leaving companion drivers on the previous
--- a/README.md
+++ b/README.md
@@ -487,15 +487,20 @@ the matching read-only Device Info properties.
 | Cover         | `CONTACT_SENSOR`                | Open/closed state contacts                                        |
 | Cover         | `RELAY`                         | Open/close/stop control relays                                    |
 | Button        | `BUTTON_LINK`                   | Allows other devices to trigger button                            |
+| Event         | `BUTTON_LINK`                   | One connection per event type, triggers other devices             |
 | Climate       | `ESPHOME_CLIMATE`               | Bind to ESPHome Climate sub-driver [\*\*](#climate-services-note) |
 | Fan           | `ESPHOME_FAN_N_SPEED[_REVERSE]` | Bind to ESPHome Fan sub-driver                                    |
 | Light         | `ESPHOME_LIGHT`                 | Bind to ESPHome Light sub-driver                                  |
 | Lock          | `ESPHOME_LOCK`                  | Bind to ESPHome Lock sub-driver                                   |
 | Water Heater  | `ESPHOME_CLIMATE`               | Bind to ESPHome Climate sub-driver                                |
 
-> **Note:** Sensor, Number, Select, Text, Text Sensor, Date, Time, Datetime, and
-> Event entities do not create bindings. They expose data only through variables
-> and events.
+> **Note:** Sensor, Number, Select, Text, Text Sensor, Date, Time and Datetime
+> entities do not create bindings. They expose data only through variables and
+> events.
+
+> **Note:** The Button and Event binding classes are the same but point opposite
+> ways. A Button binding lets another device press the ESPHome button; an Event
+> binding lets the ESPHome device act as a keypad button and drive a load.
 
 > **Note:** Water Heater entities share the `ESPHOME_CLIMATE` binding class and
 > are controlled via the ESPHome Climate sub-driver. Device modes (such as Eco,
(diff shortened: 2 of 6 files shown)


derek-miller:
The description findings and the stale merger note crossed with an edit: the
body was rewritten about four minutes before this comment, and the PR left
draft at the same time. The current body describes `DO_PUSH` then `DO_CLICK`,
the verification is an end-to-end run of the pair, and the template note and
DRV-120 line are gone.

The scan gap was real. Reproduced it first: with `DO_RELEASE` added to
`src/esphome/entities/event.lua`, `test_button_link_protocol.lua` passed, 46/0.
c69b24f extends the scan to every Lua file under `src/` and adds a "read the src
modules" check so a moved tree can't pass it silently.

Extending it tripped the "every SendToProxy occurrence was read as a call"
check once: `lib/utils.lua` forwards the name as a string,
`C4Call("SendToProxy", ...)`. Quoted occurrences are now skipped, and the check
still fails on a bare `local send = SendToProxy` in `src/`.

| mutation | result |
|---|---|
| `DO_RELEASE` in `entities/event.lua` | fails, `src/esphome/entities/event.lua: sends DO_RELEASE` |
| `DO_RELEASE` in `esphome_bthome` (control) | fails, as before |
| bare `SendToProxy` reference in `src/` | fails, parse check |

Renamed "no driver sends DO_RELEASE" to "nothing sends DO_RELEASE" to match what
it now covers.

On the squash message, agreed. Proposed text for the merge dialog, in place of
the concatenated commits:

```
feat: let event entities drive loads as keypad buttons (#113)

Each event type an ESPHome event entity declares gets its own BUTTON_LINK
connection on the keypad side, so a touch button or gesture can be bound
straight to a load in Composer. A fired event sends DO_PUSH then DO_CLICK,
the pair a Control4 keypad sends for a tap.

The "nothing sends DO_RELEASE" check now also scans the src modules, since
event.lua is the first sender outside drivers/.

Closes #103.
```

derek-miller:
Took the named exception at ea4b69c. Reproduced the gap first: `C4Call("SendToProxy", 1, "DO_RELEASE", {}, "NOTIFY")` in `esphome_bthome` passed at c69b24f, 47/0.

The skip now applies only to `C4Call("SendToProxy"` in `src/lib/utils.lua`, and the comment says it's that one site.

| arm | result |
|---|---|
| unmodified head | 47 passed |
| quoted-form send in a driver | fails, `1 occurrences did not parse as SendToProxy(...)` |
| `DO_RELEASE` in `entities/event.lua` | fails, `sends DO_RELEASE` |
| bare `SendToProxy` reference in `src/` | fails, parse check |

A rename of `lib/utils.lua` would stop the skip matching and fail the parse check, so it can't pass silently either. Full suite green, format clean.

PR #114, opened 2026-09-12
chore: update to template v0.9.23

derek-miller:
`copier update` to control4-driver-template `v0.9.23`; the update's diff is exactly the template's own v0.9.22..v0.9.23 plus `.copier-answers.yml`, no local drift, no conflict markers.

What arrives, in the order it matters to this driver:

- **#58 — the five amalgams re-vendored** from their release assets (verified byte-identical to local builds of the tags; each asset + the template's stylua flags reproduces the committed file):

  | amalgam | now | was |
  |---|---|---|
  | `bitn.lua` | v0.6.4 | v0.6.3 |
  | `crypto.lua` | v0.2.4 | v0.2.2 |
  | `noiseprotocol.lua` | v0.6.6 | v0.6.4 |
  | `bthome.lua` | v0.1.7 | v0.1.5 |
  | `protobuf.lua` | v0.6.9 | v0.6.7 |

  protobuf v0.6.9 is the one that changes wire behaviour here: NaN/infinity decode correctly instead of as ~5.1e38 (FL-15), subnormals and round-half-to-even in the float codecs (FL-16), `sfixed32` sign (FL-18), negative `int32`/`enum` on LuaJIT (FL-19), stable field and map order. bitn v0.6.4 binds `string.pack` only after probing for the 5.3 dialect, which is what makes the other four safe on a controller (its lpack raised on the previous helpers; FL-20). crypto/noiseprotocol/bthome route byte packing through bitn with unchanged output.
- **#54 — first-party NaN handling** (`tofinite`, JSON sentinels in `SerializeSafe`, NaN-aware `Values:update`). Load-bearing here: ESPHome sends NaN for any float a device has not reported yet, and with a correct decoder that now reaches `Values:update`, where `existing.value ~= value` was true on every push for a NaN and rewrote persistent storage each time (measured 5 writes vs 1 on the bot's review of #58). With #54 it is 1.
- **#57 — `SensorValueParams` / `CelsiusFromParams`** (DRV-121). Not called here yet; the call sites are the per-repo half of that ticket.
- **#59** — comment correction.

Verified locally: `make check`, `make test` 517 passed / 0 failed (up from 438: the two new template suites), and every amalgam's `selftest()` on LuaJIT under this repo's `test/c4_shim.lua` (the controller's lpack-shaped `string.pack`), all passing. On the dev controller the bot ran every amalgam's selftest one driver at a time at this exact vendor set and timed a Noise handshake (~300 ms, 16% of LuaJIT's mcode budget, no degradation over 110 handshakes); those results are on control4-driver-template#58.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Changes here will be overwritten by Copier; NEVER EDIT MANUALLY
-_commit: v0.9.22
+_commit: v0.9.23
 _src_path: git@github.com:finitelabs/control4-driver-template.git
 distributions: drivercentral oss
 github_org: finitelabs
--- a/src/lib/utils.lua
+++ b/src/lib/utils.lua
@@ -845,13 +845,26 @@ function toboolean(val)
   return false
 end
 
+--- Converts a value to a finite number.
+--- `tonumber` passes NaN and infinity straight through, so a caller's `or`
+--- fallback never fires for them.
+--- @param value any The value to convert. Can be a number or a string that represents a number.
+--- @return number|nil number The number when it is finite, or `nil` otherwise.
+function tofinite(value)
+  value = tonumber(value)
+  if value == nil or value ~= value or value == math.huge or value == -math.huge then
+    return nil
+  end
+  return value
+end
+
 --- Converts a value to a valid integer.
 --- Rounds fractional numbers and validates string representations.
 --- @param value any The value to convert to an integer. Can be a number or a string that represents a number.
 --- @return integer|nil int Returns the rounded integer if the conversion is successful, or `nil` if the value cannot be converted.
 --- @overload fun(value: number): integer
 function tointeger(value)
-  value = tonumber(value)
+  value = tofinite(value)
   if value == nil then
     return nil
   end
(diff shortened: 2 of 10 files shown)


derek-miller:
Good find, and it is exactly the class DRV-122 exists for (driver code that has to change to take advantage of the libraries it just received). Two things:

1. **Approve this as-is.** It stays a pure `copier update`, no driver edits folded in, so the tree matches the template blob-for-blob; that is the property every consumer repo's update is held to. Fix forward.
2. **Open the fix as its own PR against `main`**, to merge right behind this one, refs DRV-122:
   - `src/esphome/entities/sensor.lua`: `tofinite(state.state)` and return early when nil, before `values:update` and before the memo. The judgment call is decided: an unreported sensor's variable stays unset rather than being written as `0`. Read `missing_state` too if the schema carries it; `tofinite` is the guard either way.
   - `src/esphome/entities/number.lua:27`: same shape, same fix.
   - A test that drives `SensorEntity:updated` five times with a NaN reading and asserts 0 pushes and no variable write, with the finite control at 1, plus the same for the number entity. Both must fail on `main`.
   - Leave `esphome_climate`'s local `getCelsiusFromParams` alone in that PR; replacing it with `CelsiusFromParams` is the DRV-121 per-repo half and gets its own PR.

I will review the fix PR when it is up.

derek-miller:
@svc-finitelabs following up on my instruction above: please submit the approval on this PR as-is (it stays a pure copier update) and open the sensor NaN fix as its own PR against main. The rollout is merged everywhere else and this is the last one open.

derek-miller:
Third ask, stated as plainly as I can: **submit an approving review on this PR unchanged.** The sensor finding is real and accepted, it is not a defect in this diff, and it is being fixed in a separate PR — either one you open or one I will. This is the last of eleven v0.9.23 rollout PRs still open; the other ten are merged.

PR #123, opened 2026-09-17
fix(climate): read zero-valued state fields ESPHome leaves off the wire

### LINKED ISSUE #119: Climate -» Thermostat bug(?)
The AC unit is currently operating in cool mode. The HA reports accurately, and the C4 Thermostat proxy also reports correctly. However, if I set the mode on the HA to Off, the AC will turn off, and HA reports accurately. Nevertheless, the C4 Thermostat proxy continues to report that the unit is in cool mode.
Issue is repeatable.

derek-miller:
Fixes #119. Supersedes #124, which fixed the same bug in parallel; its fan mode
handling and its wire-level test are folded in here.

A climate unit turned off outside Control4, for example with its own remote,
kept showing its last mode on the thermostat.

ESPHome's API is proto3 with implicit presence: a field holding its type's zero
is left off the wire, and `CLIMATE_MODE_OFF` is 0. `vendor/protobuf.lua` applies
no default to a missing field, so the OFF update decoded with no `mode` at all,
and `RFP.UPDATE_STATE` only acted on fields that were present. The reporter's
driver log shows it: updates alternate between `"mode":2` and no `mode` key.

## Change

A missing field is read as zero where ESPHome's serializer always writes it.
From `APIConnection::try_send_climate_state` in 2026.8.0, the reporter's version:

| field | ESPHome writes it | read as zero when missing |
|---|---|---|
| `mode` | always | always (OFF) |
| `action` | always, even for units that never track it | if `supports_action` (OFF) |
| `fan_mode` | only when the unit has one set | if `supported_fan_modes` includes On |
| `current_temperature` | when supported | if `supports_current_temperature` |
| `current_humidity` | when supported | if `supports_current_humidity` |
| `target_humidity` | when supported | if `supports_target_humidity` |

A float is dropped only when its bits are all zero, and an unknown reading is
NaN, which is sent. So on a unit that reports temperature, a missing one means
exactly 0 C, not "no sensor".

Setpoints are left as they were. The water heater path clears a sentinel target
to nil to mean unset, so a missing setpoint has to stay missing.

The fan, lock, cover and water heater paths already default missing fields to
zero; this brings the climate driver in line.

## Reconciling with #124

Both PRs defaulted `mode` and `action` the same way. Where they differed:

- **Fan mode.** #124 read a missing fan mode as On for any unit with a fan mode
  list. Because ESPHome also omits `fan_mode` when none is set, that reports On
  for a unit that doesn't offer it, like the reporter's (`[2,3,4,5]`). Here the
  default applies only when On is in the unit's list. #123 as first pushed
  didn't handle fan On at all.
- **0 C / 0% readings.** #124 left these alone, on the grounds that defaulting
  would report 0 C for a unit with no sensor. The default here is gated on the
  support flag, and an unknown reading is NaN rather than missing, so that case
  can't arise.
- **Test.** #124 drove states through the real decoder, which is the stronger
  approach, so it's the base here. One correction: the vendored encoder writes a
  zero that ESPHome skips, so the round trip only reproduced ESPHome because the
  zero fields were left out of the input by hand. `fromWire` now strips them by
  ESPHome's rule, so a case can write `mode = OFF` or a 0 C reading directly, and
  the premise that each of those decodes to nil is asserted.

`test/test_climate_zero_fields.lua`, 38 assertions, against each version:

| driver | result | what it misses |
|---|---|---|
| `main` | 7 fail | mode, action, fan On, readings |
| #124 | 4 fail | reports On for a unit without it; 0 readings |
| #123 as first pushed | 1 fail | fan On |
| this PR | 0 fail | |
| this PR, support checks removed | 6 fail | every "unit without X" case |

Each "unit without X sends nothing" case also checks that the handler actually
ran, since it returns early on an empty entity or state.

## Verification on hardware

A test controller with a host-mode mock that flips a cool-only thermostat between
COOL + fan Auto and OFF + fan On on the device side every 20s, the way a remote
does.

Before, on `main`:

```
07:25:29 RECV mode=2 action=2    SEND HVAC_MODE_CHANGED=Cool
07:25:49 RECV mode=nil action=nil      (nothing sent)
07:26:37 device OFF since 07:26:29, proxy HVAC_MODE=Cool
```

After:

```
07:53:54 RECV mode=2 fan=2        SEND HVAC_MODE_CHANGED=Cool, FAN_MODE_CHANGED=Auto
07:54:14 RECV mode=nil fan=nil    SEND HVAC_MODE_CHANGED=Off,  FAN_MODE_CHANGED=On
07:54:23 proxy HVAC_MODE=Off FAN_MODE=On
```

`make check` clean and all suites green.

## Follow-ups

- #99 carries the same `UPDATE_STATE` code and will need a rebase onto this.
- The proxy's `HVAC_STATE` stays blank even while `Cooling` and `Off` are sent.
  That's unrelated to this issue and is tracked as DRV-131.


derek-miller:
@svc-finitelabs please review

derek-miller:
@svc-finitelabs, we have closed #124, please approve if there is nothing else.

PR #130, opened 2026-09-19
docs: make the changelog consistent — one-sentence entries, grouped by area, duplication removed

derek-miller:
## Summary

The Unreleased section accumulated across about a dozen PRs, so related entries drifted apart, two changes were each described twice, and the prose style diverged from every released section. Editorial only — no entry is dropped and no behaviour claim changes.

## Style: match the house one-sentence form

The released sections state a fix in **one sentence, no terminal period**, using a semicolon to carry the "and now it does X" half:

> Fixed the HVAC state freezing on the previous value during a heat pump's defrost cycle; defrosting now reports as Heating, and unmapped climate modes/actions log a warning instead of being silently dropped

The Unreleased entries had drifted into multi-sentence paragraphs. All 22 are rewritten to match, with no fact dropped.

While there, the eight entries in **shipped** releases that were already outliers against that style are normalized too, and the three that did not open with their section's verb are recast — including one in `### Fixed` that began "Declare ESPHome light hardware capabilities…". `Initial Release` is left alone as a legitimate special case.

Result across 77 of the 78 entries: every one opens with its section's verb (`Added` / `Fixed` / `Changed`), is a single sentence, and carries no terminal period.

The opening-verb rule is not invented here — it is the template comment at the top of `CHANGELOG.md`, which defines the sections as `- Added` / `- Fixed` / `- Changed` / `- Removed`.

## Duplication

- The humidity output had a `Fixed` entry ("temperature and humidity outputs appearing as audio connections") *and* a `Changed` entry ("the humidity output has moved to a different connection") — the same change from two directions. Folded into the `Fixed` entry, which keeps the "nothing needs reconnecting" reassurance that was the only reason the `Changed` one existed.
- The temperature scale was two `Fixed` entries: that thermostats always showed Fahrenheit, and that they kept their own copy of the Celsius/Fahrenheit choice. The second is the mechanism of the first. Merged, keeping both user-visible outcomes.
- Two entries covered one symptom with different causes — shown as connected after the device went offline, and after the IP, port or credentials changed. An installer sees one symptom, so they are one entry naming both.

## Grouping

`Fixed` is ordered by area rather than by merge order: updates, connection status, climate state, climate readings, climate connections, lights, buttons, Bluetooth. `Added` leads with the climate arc (presets, scheduling, holds, vane) so it reads as one feature.

One entry had a second, unrelated bug tacked on the end — "a thermostat staying on its last mode" trailed into "…and a temperature of exactly 0°C or a humidity of 0%". That is a zero-treated-as-absent bug, so it is split out and moved beside the 5.1e38 entry, which is the same family.

## Terminology

Three names were in use for one surface — "the app", "the Control4 app", and "Navigator". Navigator is the house term and the only one used before this release. The file is also US-spelled throughout (`color` ×10, `honored`); `greying` was the lone British outlier.

## Counts

Added 5 → 5, Fixed 16 → 15 (two merges, one split), Changed 3 → 2 (humidity folded into `Fixed`).

`README.md` is regenerated rather than hand-edited — its diff is identical to `CHANGELOG.md`'s. `make check` exits 0.


derek-miller:
Fixed at `619b927`. The connection entry finding is accepted — it was a real defect and it contradicted this PR's own stated contract.

I verified it at source rather than taking the trace, and it holds in both halves. `git show 662dc88^:src/esphome/entities/light.lua:64`, fan `:71` and lock `:64` each already walked `bindings:getDynamicBindings(self.TYPE)` and sent `UPDATE_DISCONNECT`, so those three were never broken on the device-offline path. The two that were: climate's `RFP.UPDATE_DISCONNECT` answered with `ONLINE_CHANGED { STATE = false }` (`drivers/esphome_climate/driver.lua:915` pre-#104), which thermostatV2 has no handler for, and `src/esphome/entities/water_heater.lua` had no `disconnected()` at all until #108 added one — zero matches before `28a3c0b`, one after.

Taken your wording as-is, since the scope riding the cause is the point and rephrasing it would only risk reintroducing the ambiguity.

Also taken:

- **The holds entry.** You were right that dropping the availability half goes quiet on the reachability #129 actually shipped. Restored as "offered with or without a schedule and released by no scheduled event". The next-event half stays out, which was the half that only proved the work.
- **The dangling modifier.** "looking for a button" was attaching to "scanner". Now "Fixed Bluetooth proxy recovery looking for a button with \"restart\" in its name to clear a stalled scanner".
- **The tense seam.** "the same applied to switching the fan to On" is now "switching the fan to On is now shown too", so all three clauses are present tense.
- **The setpoint entry reading worse.** Agreed — "Changed climate devices ... to show one setpoint" did read as though the driver changed the devices. Now "Changed the setpoint display for climate devices that report a single target: they now show one setpoint ...", which keeps the verb lead without moving the subject away from the behaviour.
- **The body nit.** Corrected to 77 of 78.

One correction accepted against my own commit message: I justified the verb-lead rewrite by asserting that every `Added` entry starts "Added" and every `Fixed` one "Fixed". You're right that there were four counterexamples in the file when I wrote it — "Bumped", "Declare", "Improved" and "Initial Release". The template still defines the convention, and this PR brings it to 77 of 78, but the warrant as I stated it was wrong and the later commit that fixed three of those four doesn't retroactively make it true.

Two notes on your process, both worth keeping:

Re-opening the content read at `14c9b29` instead of holding to "no second pass needed" was the right call — that commit rewrote 30 entries, 8 of them in shipped releases, and a stale approval would have covered text nobody had audited.

And checking `README.md` by running `make docs-readme` and asserting a clean `git status`, rather than inferring it from the two diffs matching, is the check that actually bites. `check-md` only asserts `mdformat --check --wrap 80`, and `make build` regenerates the README mid-build, so a hand-edited one is invisible to every gate. That's the same reasoning that found the index looseness in `754181e`: the generated artifact is where the defect lives, not the source.

Gates at this head: `make check` 0, `make docs-readme` leaves the tree clean, 78 entries with 0 terminal periods, 0 interior sentence breaks, and `Initial Release` the sole deliberate verb exception.

